PR #2, opened 2018-06-13
Transaction flex trit

ifullgaz:
Rewrite of transaction to support flex trit

PR #3, opened 2018-06-13
Transfer model

ifullgaz:
Transfer model

Depends on #2


th0br0:
@ifullgaz could you rebase this one as well please

common/model/transfer.h
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://gitlab.com/iota-foundation/software/entangled
```
jserv: Ditto.

jserv (commented):
Fix inconsistent Gitlab link

common/model/tests/BUILD
```diff
@@ -6,3 +6,13 @@ cc_test(
         "@unity",
     ],
 )
+
+cc_test (
+    name = "test_transfer",
+    srcs = ["test_transfer.c"],
+    linkopts = ["-lpthread"],
```
th0br0: Why `-lpthread`?

common/model/tests/test_transfer.c
```diff
@@ -0,0 +1,39 @@
...
+#include <string.h>
+#include <stdlib.h>
+#include <unity/unity.h>
+
+#include "common/model/transfer.h"
+
+void test_bundle_hash(coid) {
+  transfer_t transfer = transfer_new(VALUE_IN);
+  transfer_value_in_t value_in = transfer_value_in(transfer);
+  transfer_value_in_set_address(value_in, (const tryte_t *)"9XV9RJGFJJZWITDPKSQXRTHCKJAIZZY9BYLBEQUXUNCLITRQDR9CCD99AANMXYEKD9GLJGVB9HIAGRIBQ");
```
th0br0: please run `clang-format` - did you install the hooks?

common/model/tests/test_transfer.c
```diff
@@ -0,0 +1,39 @@
...
+  transfer_set_tag(transfer, (tryte_t *)"ZTDIDNQDJZGUQKOWJ9JZRCKOVGP");
```
th0br0: either cast to const here as well or don't at all?

common/model/transaction.c
```diff
@@ -23,96 +23,89 @@
...
-  offset += sizeof(transaction->obsolete_tag);
-  transaction_set_timestamp(transaction, trytes_to_long(trytes + offset, 9));
-  offset += 9;
+  tryte_t * position = (tryte_t *)trytes;
+  transaction_set_signature(transaction, position);
+  position += sizeof(transaction->signature_or_message);
+  transaction_set_address(transaction, position);
+  position += sizeof(transaction->address);
+  transaction_set_value(transaction, trytes_to_long(position, 27));
+  position += 27;
```
th0br0: could we get rid of the magic numbers here?

common/model/transaction.c
```diff
@@ -23,96 +23,89 @@
...
+  position += sizeof(transaction->signature_or_message);
+  memcpy(position, transaction_address(transaction),
          sizeof(transaction->address));
-  offset += sizeof(transaction->address);
-  _convert_and_pad(transaction_value(transaction), trytes + offset, 27);
-  offset += 27;
-  memcpy(trytes + offset, transaction_obsolete_tag(transaction),
+  position += sizeof(transaction->address);
+  long_to_trytes(transaction_value(transaction), position);
+  position += 27;
```
th0br0: magic numbers ...

th0br0 (changes requested):
Rebase and fix against master please

common/helpers/BUILD
```diff
@@ -19,6 +19,49 @@ cc_library(
         "//common/kerl",
         "//common/sign/v1:iss_kerl",
         "//common/trinary:trit_tryte",
+        "//common/trinary:trit_array",
+    ],
+)
+
+cc_library(
+    name = "sign_3",
```
th0br0: you don't need all these targets, just depend on `trit_array` directly?

common/model/transfer.c
```diff
@@ -0,0 +1,715 @@
...
+
+// Get the seed
+flex_trit_t *transfer_output_seed(transfer_output_t transfer_output) {
+  return transfer_output->seed;
+}
+
+// Set the seed (copy argument)
+void transfer_output_set_seed(transfer_output_t transfer_output,
+                              const flex_trit_t *seed) {
+  memcpy(transfer_output->seed, seed, sizeof(transfer_output->seed));
```
th0br0: We shouldn't copy here

common/model/transfer.c
```diff
@@ -0,0 +1,715 @@
...
+#define TRYTES_PER_BUNDLE_HASH 81
+
+/***********************************************************************************************************
+ * Transfer Input data structure
+ ***********************************************************************************************************/
+// To obfuscate the model, the structure definition comes here
+// struct _transfer_output {...};
+
+// Creates and returns a new transfer output
+transfer_output_t transfer_output_new(void) {
```
th0br0: I think you confused the output/input naming

common/model/transfer.c
```diff
@@ -0,0 +1,715 @@
...
+}
+
+// Set the security level
+void transfer_output_set_security(transfer_output_t transfer_output,
+                                  size_t security) {
+  transfer_output->security = security;
+}
+
+// Get the index
+size_t transfer_output_index(transfer_output_t transfer_output) {
```
th0br0: let's call this `seed_index`

common/model/transfer.c
th0br0: this is for a `value_in`(put)

common/model/transfer.c
```diff
@@ -0,0 +1,715 @@
...
+          transfer_value_out(transfer));
+    case VALUE_IN:
+      return transfer_value_in_transactions_count(transfer_value_in(transfer));
+  }
+  // Should not reach here
+  return 0;
+}
+
+// Get the tag
+flex_trit_t *transfer_tag(transfer_t transfer) { return transfer->tag; }
```
th0br0: the tag should be per part of the transfer, not global?

common/model/transfer.c
```diff
@@ -0,0 +1,715 @@
...
+// Gets the count of transactions for the transfers
+int transfer_ctx_count(transfer_ctx_t transfer_ctx) {
+  return transfer_ctx->count;
+}
+
+// Calculates the bundle hash for a collection of transfers
+int transfer_ctx_hash(transfer_ctx_t transfer_ctx, Kerl *kerl,
+                      transfer_t *transfers, size_t len) {
+  size_t i, j, count, current_index = 0;
+  tryte_t essence[TRYTES_PER_ESSENCE];
```
th0br0, line 430: why do you need tryte here?

th0br0 (commented):
Transfers should use as little memory as possible most of the time. 

Could you change the approach so that the only thing that the transfer objects keep around is pointers to the actual addresses/seeds/...?

th0br0 (changes requested):
Transfers should use as little memory as possible most of the time. 

Could you change the approach so that the only thing that the transfer objects keep around is pointers to the actual addresses/seeds/...?

common/helpers/BUILD
ifullgaz: It allows testing the signing functions with each memory model.

common/model/transfer.c
ifullgaz: transfer_output_t defines seed as an array of 243 flex trits. seed cannot be directly assigned to transfer_output->seed, it has can only be copied into the array.
With a new approach to keep pointers to data, it is possible to assign instead of copy.

common/model/transfer.c
ifullgaz: Following a discussion we had a few months ago, we agreed that the naming for in and out should be "swapped".

common/model/transfer.c
ifullgaz: I prefer not to use globals.

common/model/transfer.c
ifullgaz: In the 4:1 and 5:1 models, we would need to use the 'insert' function which would be far less efficient. When kerl support flex_trit natively, it will become advantageous since we can remove the conversion from trytes (or flex_trits) to trits also.

common/model/transfer.c
```diff
@@ -0,0 +1,673 @@
...
+#include "common/model/transfer.h"
+#include <string.h>
+#include "common/helpers/sign.h"
+#include "common/trinary/trit_tryte.h"
+#include "common/trinary/tryte_long.h"
+
+#define TRITS_PER_MESSAGE 6561
+#define TRITS_PER_ESSENCE 486
+#define TRITS_PER_BUNDLE_HASH 243
+#define TRYTES_PER_MESSAGE 2187
```
th0br0: It appears you're using these as offset multiplicators; however, you're using flextrits! That won't work.

common/model/tryte_transaction.cc
```diff
@@ -50,7 +50,7 @@ std::string TryteTransaction::message(void) {
   return trytes;
 }
 
-// Set the transaction message (copy argument)
+// Set the transaction message (assign)
 void TryteTransaction::setMessage(const std::string &trytes) {
   size_t flex_len = flex_trits_num_for_trits(NUM_TRITS_SIGNATURE);
   flex_trit_t trits[flex_len];
```
th0br0: All that the `transaction_set_message` methods do is memcpy.
Can't you just memset & copy straight into the transaction's structure? 
All of these methods waste memory

th0br0 (changes requested):

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
   return (char*)signature;
 }
 
+IOTA_EXPORT flex_trit_t* iota_flex_sign_address_gen(
+    flex_trit_t const* const seed, size_t const index, size_t const security) {
+  if (!(security > 0 && security <= 3)) return NULL;
```
thibault-martinez: add brackets

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+  const size_t key_length = security * ISS_KEY_LENGTH;
+
+  Kerl kerl;
+
+  trit_t subseed[HASH_LENGTH];
+  trit_t key[key_length];
```
thibault-martinez: use calloc for variable length static arrays (for windows)

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+  iss_kerl_key(subseed, key, key_length, &kerl);
+  memset(subseed, 0, HASH_LENGTH * sizeof(trit_t));
+
+  iss_kerl_key_digest(key, key, key_length, &kerl);
+  iss_kerl_address(key, key, security * HASH_LENGTH, &kerl);
+
+  kerl_reset(&kerl);
+
+  int8_to_flex_trit_array(address, HASH_LENGTH, key, key_length, HASH_LENGTH);
+  memset(key, 0, key_length * sizeof(trit_t));
```
thibault-martinez: Is it really useful ?

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+    flex_trit_t const* const seed, size_t const index, size_t const security) {
+  if (!(security > 0 && security <= 3)) return NULL;
+  const size_t key_length = security * ISS_KEY_LENGTH;
+
+  Kerl kerl;
+
+  trit_t subseed[HASH_LENGTH];
+  trit_t key[key_length];
+  flex_trit_t* address =
+      calloc(flex_trits_num_for_trits(HASH_LENGTH) + 1, sizeof(flex_trit_t));
```
thibault-martinez: return NULL if out of memory

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+
+  int8_to_flex_trit_array(address, HASH_LENGTH, key, key_length, HASH_LENGTH);
+  memset(key, 0, key_length * sizeof(trit_t));
+
+  return address;
+}
+
+IOTA_EXPORT flex_trit_t* iota_flex_sign_signature_gen(
+    flex_trit_t const* const seed, size_t const index, size_t const security,
+    flex_trit_t const* const bundleHash) {
```
thibault-martinez: bundle_hash

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+  const size_t key_length = security * ISS_KEY_LENGTH;
+
+  Kerl kerl;
+
+  trit_t hash[HASH_LENGTH];
+  trit_t subseed[HASH_LENGTH];
+  trit_t key[key_length];
```
thibault-martinez: calloc

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+  flex_trit_t* signature =
+      calloc(flex_trits_num_for_trits(key_length) + 1, sizeof(flex_trit_t));
```
thibault-martinez: return NULL if out of memory

common/helpers/sign.c
```diff
@@ -72,5 +73,66 @@ char* iota_sign_signature_gen(char const* const seed, size_t const index,
...
+  flex_trit_array_to_int8(hash, HASH_LENGTH, bundleHash, HASH_LENGTH,
+                          HASH_LENGTH);
+
+  iss_kerl_subseed(subseed, subseed, index, &kerl);
+  iss_kerl_key(subseed, key, key_length, &kerl);
+  memset(subseed, 0, HASH_LENGTH * sizeof(trit_t));
+
+  iss_kerl_signature(key, hash, key, key_length, &kerl);
+  int8_to_flex_trit_array(signature, key_length, key, key_length, key_length);
+  memset(key, 0, key_length * sizeof(trit_t));
```
thibault-martinez: useful ?

common/model/transaction.c
```diff
@@ -414,6 +414,22 @@ void transaction_set_hash(iota_transaction_t transaction,
...
+  memset(transaction->obsolete_tag, FLEX_TRIT_NULL_VALUE,
+         sizeof(transaction->obsolete_tag));
+  memset(transaction->bundle, FLEX_TRIT_NULL_VALUE,
+         sizeof(transaction->bundle));
+  memset(transaction->trunk, FLEX_TRIT_NULL_VALUE, sizeof(transaction->trunk));
+  memset(transaction->branch, FLEX_TRIT_NULL_VALUE,
+         sizeof(transaction->branch));
+  memset(transaction->tag, FLEX_TRIT_NULL_VALUE, sizeof(transaction->tag));
+  memset(transaction->nonce, FLEX_TRIT_NULL_VALUE, sizeof(transaction->nonce));
+}
```
thibault-martinez, line 420: What about the address and hash ?

thibault-martinez (changes requested):

common/model/transaction.c
ifullgaz: Following a discussion with Andreas, it was decided to leave them untouched.

common/helpers/sign.c
ifullgaz: I remember reading somewhere that it's good practice, to avoid "guessing" the key from the "noise" generated by the function. It was there before, I've left it untouched.

common/helpers/sign.c
ifullgaz: I remember reading somewhere that it's good practice, to avoid "guessing" the key from the "noise" generated by the function. It was there before, I've left it untouched.

tsvisabo (approved):

common/helpers/sign.c
thibault-martinez: Interesting, thanks

thibault-martinez (approved):

PR #4, opened 2018-06-14
MAM

thibault-martinez:
Added and/or updated/refactored

- MAM
- mask and unmask
- merkle tree
- encode/decode algorithms  

Notable changes

- Avoid copies for each rounds in curl_p
- Added tests for MAM, mask/unmask, merkle tree, encode/decode
- Added signed_window is ISS v1
- Avoid modification of the seed in ISS subseed


common/mam/mask.c
```diff
@@ -0,0 +1,32 @@
+#include <string.h>
+
+#include "common/trinary/add.h"
+#include "mam/mask.h"
+
+int mask(trit_t *dest, trit_t *message, size_t l, Curl *const c) {
```
th0br0: message is const

common/mam/mask.c
```diff
@@ -0,0 +1,32 @@
...
+  size_t chunk_length;
+  trit_t chunk[HASH_LENGTH];
+  for (size_t i = 0; i < l; i += HASH_LENGTH) {
+    chunk_length = l - i < HASH_LENGTH ? l - i : HASH_LENGTH;
+    memcpy(chunk, &message[i], chunk_length * sizeof(trit_t));
+    for (size_t j = 0; j < chunk_length; j++) {
+      dest[i + j] = trit_sum(chunk[j], c->state[j]);
```
th0br0: can't you just use `message[i+j]` here and remove the local `chunk` buffer?

common/mam/mask.c
```diff
@@ -0,0 +1,32 @@
...
+}
+
+int unmask(trit_t *dest, trit_t *cipher, size_t l, Curl *const c) {
+  size_t chunk_length;
+  trit_t chunk[HASH_LENGTH];
+  for (size_t i = 0; i < l; i += HASH_LENGTH) {
+    chunk_length = l - i < HASH_LENGTH ? l - i : HASH_LENGTH;
+    memcpy(chunk, &cipher[i], chunk_length * sizeof(trit_t));
+    for (size_t j = 0; j < chunk_length; j++) {
+      dest[i + j] = trit_sum(chunk[j], -c->state[j]);
```
th0br0: same here, just use `cipher[i+j]`

common/mam/mam.c
```diff
@@ -0,0 +1,187 @@
...
+                   sibling_number * HASH_LENGTH;
+  mask(payload + offset - to_mask, payload + offset - to_mask, to_mask,
+       enc_curl);
+
+  curl_reset(enc_curl);
+
+  return payload_min_length;
+}
+
+int mam_parse(trit_t *payload, size_t payload_length, trit_t *message,
```
th0br0: could you mark things as const / nonconst here?

th0br0 (changes requested):
Only did a cursory read so far;
could you also add a test or change the existing test to a non-all-9 sidekey?

common/mam/mask.c
thibault-martinez: It would remove the ability to mask in-place

common/mam/mask.c
thibault-martinez: I think it's ok for unmask though

common/sign/v1/iss.c.inc
```diff
@@ -105,7 +121,7 @@ int _ISS_PREFIX(sig_digest)(trit_t *const dig, trit_t *const hash, trit_t *sig,
     }
   }
   _HASH_PREFIX(absorb)(state, sig_start, sig_len);
-  _HASH_PREFIX(squeeze)(state, dig, HASH_LENGTH);
+  //_HASH_PREFIX(squeeze)(state, dig, HASH_LENGTH);
```
thibault-martinez: I'm not 100% sure about commenting that out as sig_digest is not tested.
Should this squeeze really be here ?
I haven't been able to complete an address from a signature digest with it being here.

common/curl-p/curl_p.c
```diff
@@ -45,9 +45,12 @@ void transform(Curl *const ctx) {
   Curl s;
   size_t round = 0;
   for (; round < ctx->type; ++round) {
-    memcpy(s.state, ctx->state, sizeof(ctx->state));
-    sbox(ctx, &s);
+    if (round % 2 == 0)
```
jserv: You can apply the trick of bitwise operations as following:
```C
if (number & 1)
    number is odd
else
    number is even
```

So that, rewrite to avoid heavy `%` operator as following:
```C
if (round & 1)
    sbox(ctx, &s);
else
    sbox(&s, cyx);
```

common/curl-p/curl_p.c
```diff
@@ -45,9 +45,12 @@ void transform(Curl *const ctx) {
...
+      sbox(&s, ctx);
+    else
+      sbox(ctx, &s);
   }
+  if (round % 2 == 1) memcpy(ctx->state, s.state, sizeof(ctx->state));
```
jserv: Ditto.
Simply, rewrite as following:
```C
if (round & 1) memcpy(ctx->state, s.state, sizeof(ctx->state));
```

common/curl-p/curl_p.c
thibault-martinez: Will do, thanks.

common/curl-p/curl_p.c
thibault-martinez: Idem.

common/mam/merkle.c
```diff
@@ -0,0 +1,195 @@
...
+
+  while (leaf_count >= 2) {
+    acc += leaf_count;
+    leaf_count = (leaf_count + 1) >> 1;
+  }
+  return acc;
+}
+
+size_t merkle_depth(size_t node_count) {
+  size_t depth = 1;
```
tsvisabo: In the edge case of `node_count` = 1 `depth` = 0, should start from 0
(i.e, returned depth should be 1 rather than 2)

common/mam/merkle.c
```diff
@@ -0,0 +1,195 @@
...
+  while (depth < tree_depth) {
+    if (width != 0) idx = idx * 2 - 1;
+    acc--;
+    depth++;
+  }
+  return acc + merkle_node_base_index(0, tree_depth, idx, base_size);
+}
+
+size_t merkle_size(size_t leaf_count) {
+  size_t acc = 1;
```
tsvisabo: here you could just return `leaf_count*2 - 1`
that's because:
pow(2,n) =sum(pow(2,n-1)) + 1
https://math.stackexchange.com/questions/1990137/the-idea-behind-the-sum-of-powers-of-2 
(assuming it is a complete tree)

common/mam/merkle.c
```diff
@@ -0,0 +1,195 @@
...
+}
+
+static size_t binary_tree_size(size_t depth, size_t acc) {
+  if (depth == 0) {
+    return acc + 1;
+  }
+  return binary_tree_size(depth - 1, acc + (1 << depth));
+}
+
+static size_t merkle_node_base_index(size_t acc, size_t depth, size_t width,
```
tsvisabo: could you add some documentation?
is acc the size of the tree in every context?

common/mam/merkle.c
```diff
@@ -0,0 +1,195 @@
...
+int merkle_create(trit_t *const tree, trit_t *const seed, int64_t offset,
+                  const size_t base_size, size_t key_size, Curl *const c) {
+  size_t depth_i, e;
+  ssize_t j;
+  trit_t key[key_size];
+
+  const size_t td = ceil_log2(base_size);
+
+  // create addresses
+  for (depth_i = 0; depth_i < base_size; depth_i++) {
```
tsvisabo: depth_i == leaf_index ?
if so, it's clearer to rename to something indicating we are traversing horizontally rather than vertically..

tsvisabo (changes requested):

common/mam/merkle.c
thibault-martinez: I think you are reviewing an outdated commit, this function does not exist anymore.

common/mam/merkle.c
thibault-martinez: I was indeed assuming it could be an incomplete tree.

common/mam/merkle.c
thibault-martinez: This has been fixed in a recent commit.

common/mam/merkle.c
thibault-martinez: Will do, thanks.

tsvisabo (approved):

common/mam/v1/mam.c
```diff
@@ -0,0 +1,188 @@
...
+  // get security back from state
+  if (offset >= payload_length) return -1;
+  trit_t hash[HASH_LENGTH];
+  memcpy(hash, enc_curl->state, HASH_LENGTH * sizeof(trit_t));
+  *security = signed_window(hash);
+  unmask(payload + offset, payload + offset, payload_length - offset, enc_curl);
+
+  if (*security == 0)
+    return -1;
+  else {
```
jserv: Here `else` statement is not necessary since Line #155 already returns if `*security` equals to zero.

common/mam/v1/mam.c
```diff
@@ -0,0 +1,188 @@
...
+      merkle_root(hash, payload + offset, siblings_number, *index, enc_curl);
+    }
+
+    // check merkle root with the given root
+    if (memcmp(hash, root, HASH_LENGTH * sizeof(trit_t)) == 0)
+      return 0;
+    else
+      return -1;
+  }
+  return 0;
```
jserv: Ditto. The statement `return 0` is not reachable.

common/mam/v1/mam.c
thibault-martinez: Indeed, thanks

common/mam/v1/BUILD
```diff
@@ -0,0 +1,32 @@
+package(default_visibility=["//visibility:public"])
+
+cc_library(name="mam",
```
th0br0: Could we move mam to `/mam` ?

common/mam/v1/BUILD
```diff
@@ -0,0 +1,32 @@
...
+           srcs=["mam.c"],
+           hdrs=["mam.h"],
+           include_prefix="mam",
```
th0br0: let's not specify a specific include prefix here so that people have to use the versioned path by default

common/mam/v1/BUILD
```diff
@@ -0,0 +1,32 @@
...
+            ":merkle",
+            "//common/curl-p:hamming",
+            "//common/trinary:trit_long",
+           ]
+)
+
+cc_library(name="mask",
+           srcs=["mask.c"],
+           hdrs=["mask.h"],
+           include_prefix="mam",
```
th0br0: same here

common/mam/v1/BUILD
```diff
@@ -0,0 +1,32 @@
...
+           [
+            "//common/curl-p:trit",
+            "//common/trinary:add"
+           ]
+)
+
+cc_library(name="merkle",
+           srcs=["merkle.c"],
+           hdrs=["merkle.h"],
+           include_prefix="mam",
```
th0br0: and here

common/mam/v1/merkle.h
```diff
@@ -0,0 +1,24 @@
...
+#define NULL_TREE 2
+#define LEAF_INDEX_OUT_OF_BOUNDS 3
+#define DEPTH_OUT_OF_BOUNDS 4
+
+size_t merkle_size(size_t leaf_count);
+size_t merkle_depth(size_t node_count);
+size_t merkle_node_index(size_t depth, size_t width, size_t tree_depth);
+size_t merkle_leaf_index(size_t leaf_index, size_t leaf_count);
+
+int merkle_create(trit_t *const tree, trit_t *const seed, int64_t offset,
```
th0br0: Could you add documentation here? Just steal from https://github.com/iotaledger/MAM/blob/master/mam/src/mam.rs

common/mam/v1/merkle.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include "common/curl-p/trit.h"
+
+#define NULL_SIBLING 1
+#define NULL_TREE 2
+#define LEAF_INDEX_OUT_OF_BOUNDS 3
+#define DEPTH_OUT_OF_BOUNDS 4
+
+size_t merkle_size(size_t leaf_count);
+size_t merkle_depth(size_t node_count);
```
th0br0: Also for this & other methods, could you add `const` qualifiers where possible?

common/mam/v1/mam.c
```diff
@@ -0,0 +1,184 @@
...
+
+  if (*security == 0) return -1;
+
+  // decrypt signature from payload
+  curl_reset(enc_curl);
+  iss_curl_sig_digest(hash, hash, 0, payload + offset,
+                      *security * ISS_KEY_LENGTH, enc_curl);
+
+  // complete the address
+  curl_reset(enc_curl);
```
th0br0: the `iss_*` methods reset curl before returning, these resets aren't necessary.

th0br0 (changes requested):

common/trinary/trit_long.c
```diff
@@ -39,60 +46,72 @@ size_t long_to_trits(int64_t value, trit_t *out) {
...
+  return value + RADIX - rem;
 }
 
 size_t encoded_length(int64_t value) {
-  size_t start, power;
-  return encoded_length_variables(value, &start, &power);
+  if (value == 0) return sizeof(encoded_zero) / sizeof(trit_t);
+  size_t length = nearest_greater_multiple_of_three(min_trits(llabs(value)));
+  // trits length + encoding length
+  return length + min_trits(pow(2, length / RADIX) - 1);
```
jserv: A left shift by n bits is defined as multiplication with pow(2,n). Rewriting with `1 << (length / RADIX)` would be slightly concise.

common/trinary/trit_long.c
```diff
@@ -39,60 +46,72 @@ size_t long_to_trits(int64_t value, trit_t *out) {
...
-  if (trits[i] == -1) {
-    size *= 3;
+  int64_t value = 0;
+  size_t encoding_start = 0;
+  while (encoding_start < length &&
+         trits_to_long(&trits[encoding_start], RADIX) < 0)
+    encoding_start += RADIX;
+  if (encoding_start >= length) return -1;
+  encoding_start += RADIX;
+  size_t encoding_length = min_trits(pow(2, encoding_start / RADIX) - 1);
```
jserv: Ditto. Use left shift.

common/trinary/trit_long.c
```diff
@@ -39,60 +46,72 @@ size_t long_to_trits(int64_t value, trit_t *out) {
...
+  size_t encoding = trits_to_long(&trits[encoding_start], encoding_length);
+  for (size_t i = 0; i < encoding_start / RADIX; i += 1) {
+    int64_t tryte_value = trits_to_long(&trits[i * RADIX], RADIX);
+    if ((encoding >> i) & 1) tryte_value = -tryte_value;
+    value += pow(27, i) * tryte_value;
```
jserv: Since variable `i` might only equal to limited possible numbers, would you consider to construct a dedicated lookup table to avoid duplicated computation caused by `pow`?

thibault-martinez: Thanks, lookup table for this is actually ~18% faster than pow function !

th0br0 (approved):

PR #6, opened 2018-06-16
tools/ci: Bump Bazel version

jserv:


th0br0:
Thanks, but we no longer use these scripts for CI. They'll be removed soon.

PR #36, opened 2018-07-01
cppclient - check for more null cases and return optional in getNodeInfo

tsvisabo:
Return optional on getNodeInfo
treat other null cases

### DIFF
--- a/cppclient/api_json.cc
+++ b/cppclient/api_json.cc
@@ -37,12 +37,12 @@ DEFINE_uint32(maxNumAddressesForGetBalances, 1000,
 namespace cppclient {
 
 bool IotaJsonAPI::isNodeSolid() {
-  auto niOptional = getNodeInfo();
-  if (!niOptional.has_value()) {
+  auto maybeNodeInfo = getNodeInfo();
+  if (!maybeNodeInfo.has_value()) {
     LOG(INFO) << __FUNCTION__ << " request failed.";
-    return {};
+    return false;
   }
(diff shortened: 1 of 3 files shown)


cppclient/api_json.cc
```diff
@@ -30,8 +30,12 @@ DEFINE_uint32(maxNumAddressesForGetBalances, 1000,
 namespace cppclient {
 
 bool IotaJsonAPI::isNodeSolid() {
-  auto ni = getNodeInfo();
-
+  auto niOptional = getNodeInfo();
```
th0br0: let's call these `maybeNodeInfo`?

cppclient/api_json.cc
```diff
@@ -30,8 +30,12 @@ DEFINE_uint32(maxNumAddressesForGetBalances, 1000,
...
+  if (!niOptional.has_value()) {
+    LOG(INFO) << __FUNCTION__ << " request failed.";
+    return {};
```
th0br0: this should return `false` explicitly

th0br0 (changes requested):

cppclient/api_json.cc
tsvisabo: definitley

cppclient/api_json.cc
tsvisabo: maybe :)

th0br0 (approved):

th0br0:
Rebase on master please

PR #37, opened 2018-07-01
Network components API headers

thibault-martinez:


common/network/queues/concurrent_queue_trytes.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __COMMON_NETWORK_CONCURRENT_QUEUE_TRYTES_H__
+#define __COMMON_NETWORK_CONCURRENT_QUEUE_TRYTES_H__
+
+#include "common/trinary/tryte.h"
+
+#include "concurrent_queue.h.inc"
+
+typedef tryte_t* tryte_p;
+
+DECLARE_CONCURRENT_QUEUE_OF(tryte_p);
```
th0br0: Let's use `flex_trit_p` here?

common/network/components/broadcaster.h
```diff
@@ -0,0 +1,11 @@
+#ifndef __COMMON_NETWORK_COMPONENTS_BROADCASTER_H__
+#define __COMMON_NETWORK_COMPONENTS_BROADCASTER_H__
+
+#include "common/network/queues/concurrent_queue_trytes.h"
+
+void broadcaster_on_next(tryte_t const *const hash);
```
th0br0: should also use `flex_trit_p` here

common/network/components/broadcaster.h
```diff
@@ -0,0 +1,11 @@
...
+void *broadcaster_routine(void *arg);
+
+extern concurrent_queue_of_tryte_p broadcaster_queue;
```
th0br0: marking these as extern here means that we won't be able to have multiple instances in parallel.
context/state should rather be the first argument to the method

common/network/components/processor.h
```diff
@@ -0,0 +1,11 @@
+#ifndef __COMMON_NETWORK_COMPONENTS_PROCESSOR_H__
+#define __COMMON_NETWORK_COMPONENTS_PROCESSOR_H__
+
+#include "common/network/queues/concurrent_queue_trytes.h"
+
+void processor_on_next(tryte_t const *const hash);
+void *processor_routine(void *arg);
+
+extern concurrent_queue_of_tryte_p processor_queue;
```
th0br0: same here

common/network/components/processor.h
```diff
@@ -0,0 +1,11 @@
+#ifndef __COMMON_NETWORK_COMPONENTS_PROCESSOR_H__
+#define __COMMON_NETWORK_COMPONENTS_PROCESSOR_H__
+
+#include "common/network/queues/concurrent_queue_trytes.h"
+
+void processor_on_next(tryte_t const *const hash);
```
th0br0: same here

common/network/components/receiver.h
```diff
@@ -0,0 +1,16 @@
...
+#define __COMMON_NETWORK_COMPONENTS_RECEIVER_H__
+
+#include <stdint.h>
+
+typedef struct {
+  uint16_t port;
+  int socket;
+  int packet_size;
+  void *packet;
+} receiver_ctx_t;
```
th0br0: could you add some description of what this is supposed to be used for etc?

common/network/components/requester.h
```diff
@@ -0,0 +1,11 @@
+#ifndef __COMMON_NETWORK_COMPONENTS_REQUESTER_H__
+#define __COMMON_NETWORK_COMPONENTS_REQUESTER_H__
+
+#include "common/network/queues/concurrent_queue_trytes.h"
+
+void request_transaction(tryte_t *const hash);
```
th0br0: same

common/network/components/requester.h
```diff
@@ -0,0 +1,11 @@
...
+tryte_t *get_transaction_to_request();
+
+extern concurrent_queue_of_tryte_p requester_queue;
```
th0br0: same

common/network/components/responder.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __COMMON_NETWORK_COMPONENTS_RESPONDER_H__
+#define __COMMON_NETWORK_COMPONENTS_RESPONDER_H__
+
+#include "common/network/queues/concurrent_queue_hash_request.h"
+#include "common/trinary/tryte.h"
+
+void responder_on_next(tryte_t const *const hash, neighbor_t *const neighbor);
```
th0br0: use `flex_trit_p` here instead of `tryte_t`

common/network/components/responder.h
```diff
@@ -0,0 +1,12 @@
...
+void *responder_routine(void *arg);
+
+extern concurrent_queue_of_hash_request_t responder_queue;
```
th0br0: same

common/network/neighbor.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __COMMON_MODEL_NEIGHBOR_H__
+#define __COMMON_MODEL_NEIGHBOR_H__
+
+typedef struct {
+  char *address;
```
th0br0: Maybe a different name?

th0br0 (changes requested):

common/network/concurrent_queue.h
```diff
@@ -0,0 +1,89 @@
+#include <pthread.h>
+#include <stdlib.h>
+
+#define DECLARE_STRUCT_CONCURRENT_QUEUE_NODE_OF(TYPE)  \
+  typedef struct concurrent_queue_node_of_##TYPE##_s { \
+    TYPE data;                                         \
+    struct concurrent_queue_node_of_##TYPE##_s *next;  \
+  } concurrent_queue_node_of_##TYPE##_t;
+
+#define DECLARE_VTABLE_CONCURRENT_QUEUE_OF(TYPE)                           \
```
tsvisabo: What's the reason for making the queue polymorphic?

jserv:
What is the consideration of filename `common/network/queues/concurrent_queue.h.inc`? Is it generated by some utilities?

common/network/concurrent_queue.h
thibault-martinez: Many queues with different content types

thibault-martinez:
Hi @jserv, are you speaking about the path or the actual filename ?

jserv:
@thibault-martinez, In particular, I meant the suffix `.h.inc`. I thought it would be a header file.

thibault-martinez:
It means that it shouldn't be used directly but rather included to "generate" a header file (e.g. https://github.com/thibault-martinez/entangled/blob/0cbbb4b65c7d1a79230fe6759006d08fc304b859/common/network/queues/concurrent_queue_trit_array.h)

common/network/concurrent_queue.h
th0br0: Instead of using a vtable struct you could just define the methods via macros though?

common/network/queues/concurrent_queue.h.inc
```diff
@@ -0,0 +1,95 @@
...
+    size_t (*size)(struct concurrent_queue_of_##TYPE##_s * this);          \
+    void (*push)(struct concurrent_queue_of_##TYPE##_s * this, TYPE data); \
+    void (*pop)(struct concurrent_queue_of_##TYPE##_s * this);             \
+  } vtable_concurrent_queue_of_##TYPE##_t;                                 \
+  extern vtable_concurrent_queue_of_##TYPE##_t const                       \
+      vtable_concurrent_queue_of_##TYPE##_g;
+
+#define DECLARE_STRUCT_CONCURRENT_QUEUE_OF(TYPE)         \
+  typedef struct concurrent_queue_of_##TYPE##_s {        \
+    concurrent_queue_node_of_##TYPE##_t *list;           \
```
th0br0: what do you need the `list` member for?

common/network/concurrent_queue.h
thibault-martinez: Yes, but they would have type-related name

common/network/queues/concurrent_queue.h.inc
```diff
@@ -0,0 +1,95 @@
...
+    concurrent_queue_node_of_##TYPE##_t *front;          \
+    concurrent_queue_node_of_##TYPE##_t *back;           \
+    vtable_concurrent_queue_of_##TYPE##_t const *vtable; \
+    size_t size;                                         \
+    pthread_rwlock_t lock;                               \
```
th0br0: pthread is not available on embedded necessarily.
let's find a sane way to abstract this?

common/network/concurrent_queue.h
th0br0: yes. is that so bad? you'll probably end up using macros for accessing the queue anyway

thibault-martinez: Not bad, less handy depending on the type. When declared and defined, there are no macros anymore. More on this : http://cecilsunkure.blogspot.com/2012/08/generic-programming-in-c.html

common/network/queues/concurrent_queue.h.inc
thibault-martinez: You're right

common/network/queues/concurrent_queue.h.inc
thibault-martinez: Done

common/network/neighbor.h
```diff
@@ -0,0 +1,29 @@
+#ifndef __COMMON_NETWORK_NEIGHBOR_H__
+#define __COMMON_NETWORK_NEIGHBOR_H__
+
+#include <stdlib.h>
```
jserv: Why would `<stdlib.h>` be included? It seems that the following statements do not rely on what `<stdlib.h>` declares.

common/network/queues/concurrent_queue.h.inc
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Returns pointer to the first element in the queue. This element will be the
+ * first element to be removed on a call to pop()
+ *
+ * @return Pointer to the first element
+ */
+#define DECLARE_FUNCTION_FRONT_CONCURRENT_QUEUE_OF(TYPE) \
+  TYPE *CONCURRENT_QUEUE_OF_##TYPE##_FRONT(              \
+      struct concurrent_queue_of_##TYPE##_s *this);
```
jserv: How about changing `this` to `thiz` in order to make C++ compilers happy?

common/network/neighbor.h
thibault-martinez: That's an oversight, thanks

common/network/queues/concurrent_queue.h.inc
thibault-martinez: I'll call it self, thanks

common/network/components/BUILD
```diff
@@ -0,0 +1,38 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "broadcaster",
+    hdrs = ["broadcaster.h"],
+    deps = [
+        "//common/network/queues:concurrent_queue_trit_array",
```
jserv, line 7: Would it look more concise to strip `/common/network/queues:concurrent_queue_trit_array` into `/common/network/queues:conqueue_trit_array` as [C++ Concurrent Queues](http://open-std.org/JTC1/SC22/WG21/docs/papers/2016/p0260r0.html) specifies the abbreviation `conqueues`?

jserv:
Should copyright notice be inserted in each header files?
```C
/*
 * Copyright (c) 2018 IOTA Stiftung
 * https://github.com/iotaledger/entangled
 *
 */
```

thibault-martinez:
@jserv I guess so, I'll add them.

By the way, @th0br0 @tsvisabo, we don't have the LICENSE file it's referring to.

common/network/components/BUILD
thibault-martinez: Why not, but it's only a matter of bazel as the actual type is typedefed down to something shorter

common/network/components/broadcaster.h
```diff
@@ -0,0 +1,19 @@
...
+#ifndef __COMMON_NETWORK_COMPONENTS_BROADCASTER_H__
+#define __COMMON_NETWORK_COMPONENTS_BROADCASTER_H__
+
+#include "common/network/queues/concurrent_queue_trit_array.h"
+
+typedef concurrent_queue_of_trit_array_p broadcaster_queue_t;
+
+void broadcaster_on_next(broadcaster_queue_t *const queue,
+                         trit_array_t const *const hash);
+void *broadcaster_routine(void *arg);
```
th0br0: all these routines/on_next should be given a `broadcaster_state_t` or similar context object so that we can run multiple in parallel if necessary

common/network/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+#include <unistd.h>
+
+#ifdef _POSIX_THREADS
+
+#include <pthread.h>
+typedef pthread_cond_t cond_handle_t;
+
+#else
+
+#error "No conditionnal variable primitives found"
```
th0br0: typo

common/network/iota_packet.h
```diff
@@ -0,0 +1,21 @@
...
+#ifndef __COMMON_NETWORK_IOTA_PACKET_H__
+#define __COMMON_NETWORK_IOTA_PACKET_H__
+
+#include "common/network/socket.h"
+#include "common/trinary/tryte.h"
+
+#define TRANSACTION_PACKET_SIZE 1650
+
+typedef struct {
+  tryte_t data[TRANSACTION_PACKET_SIZE];
```
th0br0: it's not 1650 trytes, but 8019 trits encoded as 1604 bytes + 230 trytes encoded as 46 bytes.
correct type should be char* ?

common/network/iota_packet.h
```diff
@@ -0,0 +1,21 @@
...
+  socket_addr_t source;
```
th0br0: `socket_addr_t` is POSIX as well

common/network/neighbor.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef __COMMON_NETWORK_NEIGHBOR_H__
+#define __COMMON_NETWORK_NEIGHBOR_H__
+
+#include "common/network/iota_packet.h"
+
+typedef enum connection_type_e { TCP, UDP } connection_type_t;
+
+typedef struct {
+  socket_addr_t endpoint;
```
th0br0: POSIX

common/network/neighbor.h
```diff
@@ -0,0 +1,34 @@
...
+  int socket;
```
th0br0: some targets might not understand the notion of a socket?

could we place endpoint & socket and `connection_type` in some generic `void*` blob?

common/network/neighbor.h
```diff
@@ -0,0 +1,34 @@
...
+  connection_type_t connection_type;
+  unsigned int number_of_all_transactions;
+  unsigned int number_of_new_transactions;
+  unsigned int number_of_invalid_transactions;
+  unsigned int number_of_sent_transactions;
+  unsigned int random_transaction_requests;
+  int flagged;
```
th0br0, line 20: what does `flagged` mean?

common/network/lock_handle.h
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __COMMON_NETWORK_LOCK_HANDLE_H__
```
th0br0: these locks etc. are not useful just for networking stuff.
let's put this in `common/lock_handle.h`?

th0br0: (same for cond_handle)

th0br0 (changes requested):

common/network/iota_packet.h
thibault-martinez: That's actually a type I declare in network/socket.h

common/network/neighbor.h
thibault-martinez: Took the neighbours fields from IRI, guess it means it could be removed soon

common/network/neighbor.h
thibault-martinez: The endpoint is the real host:port of the neighbor, should it be in the blob ?

common/network/components/broadcaster.h
thibault-martinez: arg would've been that, it was just to satisfy pthread_create requirements without warning.
But yes, I'll declare some state structs

th0br0: Ah :) fair enough. we can just cast the `pthread_create` part though I guess.
`on_next` will need a state though I guess.

common/network/neighbor.h
th0br0: I think we only need some obscure identifier for the neighbour. That can be the memory address (i.e. `void*`) of the description struct.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef COMMON_COND_HANDLE_H_
+#define COMMON_COND_HANDLE_H_
+
+/**
+ * We declare a type cond_handle_t that depends on the system conditional
```
jserv: The use of "condition variable" should be more common than "conditional variable".
Reference: https://en.wikipedia.org/wiki/Monitor_(synchronization)

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+#include <unistd.h>
+
+#ifdef _POSIX_THREADS
+
+#include <pthread.h>
+typedef pthread_cond_t cond_handle_t;
+
+#else
+
+#error "No conditional variable primitives found"
```
jserv: Ditto.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+
+#endif  // _POSIX_THREADS
+
+/**
+ * The following functions are defined contextually with macros depending on the
+ * system available conditional variable primitives, some of them might have
```
jserv: Ditto.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+ * no effect if not needed by the underlying API
+ */
+
+/**
+ * Initializes a condition variable object with the specified attributes for
+ * use
+ *
+ * @param cond The conditional variable
```
jserv: Ditto.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @return exit status
+ */
+int cond_handle_init(cond_handle_t* const cond);
+
+/**
+ * Wakes up at least one thread that is currently waiting on the condition
+ * variable specified by cond
+ *
+ * @param cond The conditional variable
```
jserv: Ditto.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @return exit status
+ */
+int cond_handle_signal(cond_handle_t* const cond);
+
+/**
+ * Wakes up all threads that are currently waiting on the condition variable
+ * specified by cond
+ *
+ * @param cond The conditional variable
```
jserv: Ditto.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @return exit status
+ */
+int cond_handle_broadcast(cond_handle_t* const cond);
+
+/**
+ * Blocks the calling thread, waiting for the condition specified by cond to be
+ * signaled or broadcast to
+ *
+ * @param cond The conditional variable
```
jserv: Ditto.

common/cond_handle.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @return exit status
+ */
+int cond_handle_wait(cond_handle_t* const cond);
+
+/**
+ * Destroys the condition variable specified by cond
+ *
+ * @param cond The conditional variable
```
jserv: Ditto.

jserv (commented):
Using "condition variable" would be more specific to monitor rather than "conditional variable".

common/cond_handle.h
thibault-martinez: You're right, I don't even know why I called them like that in the first place

thibault-martinez: Done, thanks

common/network/components/receiver.h
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __COMMON_NETWORK_COMPONENTS_RECEIVER_H__
+#define __COMMON_NETWORK_COMPONENTS_RECEIVER_H__
+
+void *receiver_routine(void *arg);
```
th0br0: does receiver not have a state?

th0br0 (approved):

PR #58, opened 2018-07-09
Concurrent queue / Lock handle / Cond handle

thibault-martinez:


common/cond_handle.c
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "cond_handle.h"
+
+#ifdef _POSIX_THREADS
```
jserv: Since you already raise `#error` in header `cond_handle.h` for non-Pthread case, is it worthy to duplicate error handling in C source files?

common/cond_handle.c
```diff
@@ -0,0 +1,36 @@
...
+
+int cond_handle_init(cond_handle_t* const cond) {
```
jserv: How about implementing these function as `static inline` in header files rather than C source? So that, C compiler can always perform optimizations.

thibault-martinez: Sure

common/cond_handle.c
thibault-martinez: Will be fixed by having only a header file

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+      struct concurrent_queue_of_##TYPE##_s *const self) { \
+    if (self == NULL) return NULL;                         \
+    lock_handle_lock(&self->lock);                         \
+    if (self->back == NULL) {                              \
+      lock_handle_unlock(&self->lock);                     \
+      return NULL;                                         \
+    }                                                      \
+    TYPE *data = &self->back->data;                        \
+    lock_handle_unlock(&self->lock);                       \
+    return data;                                           \
```
th0br0: Could we structure these more like

```c
TYPE *data = NULL;
if(self == NULL) goto end; // (or return NULL here)
lock_handle_lock(&self->lock);
if(self->back) data = self->back->data;
lock_handle_unlock(&self->lock);

end:
return data;
```

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+#define DEFINE_FUNCTION_EMPTY_CONCURRENT_QUEUE_OF(TYPE)    \
+  int CONCURRENT_QUEUE_OF_##TYPE##_EMPTY(                  \
+      struct concurrent_queue_of_##TYPE##_s *const self) { \
+    if (self == NULL) return 1;                            \
+    lock_handle_lock(&self->lock);                         \
+    size_t size = self->size;                              \
```
th0br0: let's define local variables at the start of the method in proper C style?

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+    lock_handle_unlock(&self->lock);                       \
+    return size;                                           \
+  }
+
+#define DEFINE_FUNCTION_PUSH_CONCURRENT_QUEUE_OF(TYPE)                      \
+  int CONCURRENT_QUEUE_OF_##TYPE##_PUSH(                                    \
+      struct concurrent_queue_of_##TYPE##_s *const self, TYPE const data) { \
+    if (self == NULL) return 1;                                             \
+    concurrent_queue_node_of_##TYPE##_t *node = malloc(sizeof(*node));      \
+    if (node == NULL) return 1;                                             \
```
th0br0: this should be a different error code than self being empty... use some enum?

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+    size_t size = self->size;                              \
+    lock_handle_unlock(&self->lock);                       \
+    return size;                                           \
+  }
+
+#define DEFINE_FUNCTION_PUSH_CONCURRENT_QUEUE_OF(TYPE)                      \
+  int CONCURRENT_QUEUE_OF_##TYPE##_PUSH(                                    \
+      struct concurrent_queue_of_##TYPE##_s *const self, TYPE const data) { \
+    if (self == NULL) return 1;                                             \
+    concurrent_queue_node_of_##TYPE##_t *node = malloc(sizeof(*node));      \
```
th0br0: use `calloc` so you don't need to set `node->next`

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+    if (self->front == NULL) {                                              \
+      self->front = node;                                                   \
+      self->back = node;                                                    \
+    } else {                                                                \
+      self->back->next = node;                                              \
+      self->back = node;                                                    \
+    }                                                                       \
+    self->size++;                                                           \
+    lock_handle_unlock(&self->lock);                                        \
+    cond_handle_signal(&self->cond);                                        \
```
th0br0, line 117: all methods lock; what exact purpose is the condition variable serving?

th0br0 (changes requested):

common/network/queues/concurrent_queue.c.inc
thibault-martinez: Much better, thanks

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+      lock_handle_unlock(&self->lock);                     \
+      return NULL;                                         \
+    }                                                      \
+    TYPE *data = &self->back->data;                        \
+    lock_handle_unlock(&self->lock);                       \
+    return data;                                           \
+  }
+
+#define DEFINE_FUNCTION_EMPTY_CONCURRENT_QUEUE_OF(TYPE)    \
+  int CONCURRENT_QUEUE_OF_##TYPE##_EMPTY(                  \
```
jserv: Would it be explicit to use `bool` defined by `<stdbool.h>` rather than `int`?

common/cond_handle.h
```diff
@@ -9,28 +9,20 @@
 #define COMMON_COND_HANDLE_H_
 
 /**
- * We declare a type cond_handle_t that depends on the system condition variable
- * primitives
+ * We define a type cond_handle_t, depending on the system condition variable
```
jserv: How about changing wording from "depending on" to "mapping"? It would behave as this commit did.

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,126 @@
...
+      vtable_concurrent_queue_of_##TYPE##_g =   \
+          {CONCURRENT_QUEUE_OF_##TYPE##_FRONT,  \
+           CONCURRENT_QUEUE_OF_##TYPE##_BACK,   \
+           CONCURRENT_QUEUE_OF_##TYPE##_EMPTY,  \
+           CONCURRENT_QUEUE_OF_##TYPE##_SIZE,   \
+           CONCURRENT_QUEUE_OF_##TYPE##_PUSH,   \
+           CONCURRENT_QUEUE_OF_##TYPE##_POP};
+
+#define NEW_CONCURRENT_QUEUE_OF(TYPE, NAME)                            \
+  struct concurrent_queue_of_##TYPE##_s *NAME = malloc(sizeof(*NAME)); \
```
jserv: Any potential memory leak resulting from this allocation?

common/network/queues/concurrent_queue.c.inc
thibault-martinez: It allows `pop()` to wait for new data without having to continuously poll

common/network/queues/concurrent_queue.c.inc
thibault-martinez: The user has the responsibility to free the queue

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,110 @@
...
+    lock_handle_unlock(&self->lock);                       \
+    return size;                                           \
+  }
+
+#define DEFINE_FUNCTION_PUSH_CONCURRENT_QUEUE_OF(TYPE)                      \
+  concurrent_queue_status_t CONCURRENT_QUEUE_OF_##TYPE##_PUSH(              \
+      struct concurrent_queue_of_##TYPE##_s *const self, TYPE const data) { \
+    concurrent_queue_node_of_##TYPE##_t *node;                              \
+    if (self == NULL) return CONCURRENT_QUEUE_NULL_SELF;                    \
+    node = calloc(1, sizeof(*node));                                        \
```
th0br0: to reduce risk of errors, rather do `sizeof(concurrent_queue_node_of_##TYPE##_t)` ? `sizeof(*node)` is a bit hard to read/parse,
`*node` effectively dereferences a nullptr

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+#define DEFINE_FUNCTION_POP_CONCURRENT_QUEUE_OF(TYPE)                        \
+  concurrent_queue_status_t CONCURRENT_QUEUE_OF_##TYPE##_POP(                \
+      struct concurrent_queue_of_##TYPE##_s *const self, TYPE *const data) { \
+    if (self == NULL) return CONCURRENT_QUEUE_NULL_SELF;                     \
+    if (data == NULL) return CONCURRENT_QUEUE_NULL_DATA;                     \
+    lock_handle_lock(&self->lock);                                           \
+    while (self->size == 0) {                                                \
+      cond_handle_wait(&self->cond, &self->lock);                            \
```
th0br0, line 128: this won't work though?
you hold the lock, so `push` won't be able to acquire the lock.

common/network/queues/concurrent_queue.c.inc
```diff
@@ -0,0 +1,110 @@
...
+    }                                                                        \
+    concurrent_queue_node_of_##TYPE##_t *tmp = self->front;                  \
+    *data = tmp->data;                                                       \
```
th0br0, line 131: this won't work? you'd need to do a memcpy of tmp->data to data?

th0br0 (changes requested):

common/network/queues/concurrent_queue.c.inc
thibault-martinez: The action to wait on a condition variable actually releases the associated lock

> The pthread_cond_wait() function atomically blocks the current thread waiting on the condition variable specified by cond, and releases the mutex
>      specified by mutex.  The waiting thread unblocks only after another thread calls pthread_cond_signal(3), or pthread_cond_broadcast(3) with the same
>      condition variable, and the current thread reacquires the lock on mutex.

th0br0: fair enough!

common/network/queues/concurrent_queue.c.inc
thibault-martinez: tmp->data is not a pointer so it's a simple copy

jserv (commented):
Abstract object allocation and de-allocation inside macro `INIT_CONCURRENT_QUEUE_OF` by introducing another macro `FINI_CONCURRENT_QUEUE_OF`. So that, developers don't have to be aware of the use of `free` and set the pointer to null.

thibault-martinez:
@jserv I initially thought about that but it ended up being only a free so I removed it. But I guess you're right on the abstraction part, I'll add it

I don't think it will need the `_OF` though

EDIT : we need to destroy the queue lock/cond so this new macro is justified. We also need to empty the queue if it is not.

common/network/queues/concurrent_queue.h.inc
```diff
@@ -5,26 +5,41 @@
  * Refer to the LICENSE file for licensing information
  */
 
+#ifndef __COMMON_NETWORK_CONCURRENT_QUEUE_H__
+#define __COMMON_NETWORK_CONCURRENT_QUEUE_H__
+
+#include <stdbool.h>
+#include <stdlib.h>
```
jserv: Why did you include `<stdlib.h>` here?

thibault-martinez: It should be in concurrent_queue.c.inc, good catch

th0br0 (approved):

Issue #60, opened 2018-07-16
Atomic handle


thibault-martinez:
Design / Implement macros to define atomic_handle_t types, that depends on a type, and its associated functions

PR #67, opened 2018-07-16
Implementation base of network components

thibault-martinez:
PS : the boolean `running` used in the components should be atomic.
In "most cases" it is natively atomic as bool are 1 byte long.

However, the standard states
`bool - type, capable of holding one of the two values: true or false. The value of sizeof(bool) is implementation defined and might differ from 1.`

We should then enforce the atomicity, research will be done in #60


common/thread_handle.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <unistd.h>
+
+#ifdef _POSIX_THREADS
+
+#include <pthread.h>
+
+typedef pthread_t thread_handle_t;
+
+/**
```
th0br0: Could we move the documentation elsewhere so it doesn't have to be duplicated for each impl? i.e. forward decl or so

common/network/components/broadcaster.c
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "broadcaster.h"
```
jserv: Why didn't you include `common/network/components/broadcaster.h` for consistency?

thibault-martinez: No particular reason, maybe same folder is smoother in case we change paths. But I agree for consistency.

common/network/components/processor.c
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+  return NULL;
+}
+
+bool processor_start(processor_state_t *const state) {
+  if (state == NULL) return false;
+  if (INIT_CONCURRENT_QUEUE_OF(trit_array_p, state->queue) !=
+      CONCURRENT_QUEUE_SUCCESS)
+    return false;
```
th0br0: let's add `{}`

common/network/components/broadcaster.c
```diff
@@ -0,0 +1,52 @@
...
+      // TODO(thibault) broadcast the hash
+    }
+  }
+  return NULL;
+}
+
+bool broadcaster_start(broadcaster_state_t *const state) {
+  if (state == NULL) return false;
+  if (INIT_CONCURRENT_QUEUE_OF(trit_array_p, state->queue) !=
+      CONCURRENT_QUEUE_SUCCESS)
```
th0br0: let's add `{}`

common/network/components/requester.c
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "requester.h"
```
th0br0: absolute include paths

common/network/components/requester.c
```diff
@@ -0,0 +1,41 @@
...
+
+bool requester_start(requester_state_t *const state) {
+  if (state == NULL) return false;
+  if (INIT_CONCURRENT_QUEUE_OF(trit_array_p, state->queue) !=
```
th0br0: add `[}`

common/network/components/requester.c
```diff
@@ -0,0 +1,41 @@
...
+  if (state == NULL) return false;
+  return state->queue->vtable->push(state->queue, hash) ==
+         CONCURRENT_QUEUE_SUCCESS;
+}
+
+trit_array_p get_transaction_to_request(requester_state_t *const state) {
+  trit_array_p hash;
+
+  if (state == NULL) return NULL;
+  if (state->queue->vtable->pop(state->queue, &hash) !=
```
th0br0: add `{}`

common/network/components/requester.c
```diff
@@ -0,0 +1,41 @@
...
+
+bool request_transaction(requester_state_t *const state,
+                         trit_array_p const hash) {
+  if (state == NULL) return false;
+  return state->queue->vtable->push(state->queue, hash) ==
+         CONCURRENT_QUEUE_SUCCESS;
+}
+
+trit_array_p get_transaction_to_request(requester_state_t *const state) {
+  trit_array_p hash;
```
th0br0, line 31: `trit_array_p` is a ptr.
`typedef struct _trit_array *trit_array_p;`

The `pop` will cause a nullptr-deref

common/network/components/processor.c
```diff
@@ -0,0 +1,52 @@
...
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "processor.h"
+#include "common/network/logger.h"
+
+static void *processor_routine(processor_state_t *const state) {
+  trit_array_p hash;
```
th0br0: nullptr deref on `pop`, `trit_array_p` is a ptr

common/network/components/processor.c
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "processor.h"
```
th0br0: absolute include paths

common/cond_handle.h
```diff
@@ -73,6 +73,23 @@ static inline int cond_handle_wait(cond_handle_t* const cond,
   return pthread_cond_wait(cond, lock);
 }
 
+/**
```
th0br0: same here, can we move the documentation to somewhere else so we don't have to redefine each time?

th0br0 (changes requested):

common/thread_handle.h
thibault-martinez: Totally agree, I've been thinking the same. Can't decl them before the actual type exists though, I can declare them separately at the end of the file

common/network/components/requester.c
thibault-martinez: Maybe the queue should store trit_array_t and not trit_array_p ?

th0br0 (approved):

PR #71, opened 2018-07-19
Fix style tweak

howjmay:
There is a style tweak in the comment in the cclient/iota_api.h.


jserv:
Change subject to `Style tweak`.

PR #72, opened 2018-07-20
Style tweak

howjmay:
There is a style tweak in the comment.


jserv:
Use `git commit --amend` to rework this commit. Since it is just minor style tweak, you don't really have to explain much.

PR #83, opened 2018-07-25
Fix readme - tanglescope metric units from ms to seconds

tsvisabo:


jserv:
The subject of this pull request did not reflect the changed files.

PR #88, opened 2018-07-26
cIRI node base and binary

thibault-martinez:


ciri/main.c
```diff
@@ -0,0 +1,31 @@
...
+
+int main() {
+  logger_init(LOG_DEBUG, false, stdout);
+
+  log_info("Starting cIRI node");
+  if (node_start(&node_g) == false) {
+    return EXIT_FAILURE;
+  }
+
+  sleep(10);
```
jserv, line 23: Why hardcode `sleep(10)` here? Can you use notification internally?

thibault-martinez: For now it's just a placeholder, it will be removed as soon as the main thread is doing something

th0br0 (approved):

PR #96, opened 2018-07-29
Add milestones counter to statscollector

tsvisabo:


tanglescope/common/iri.cpp
```diff
@@ -17,7 +17,9 @@ std::shared_ptr<IRIMessage> payloadToMsg(std::string_view payload) {
     return std::make_shared<SNMessage>(actual);
   } else if (what == "lmhs") {
     return std::make_shared<LMHSMessage>(actual);
-  } else if (what == "rstat") {
+  } else if (what == "lmsi") {
+    return std::make_shared<LMSIMessage>(actual);
+  }else if (what == "rstat") {
```
th0br0: `clang-format` please

tanglescope/common/iri.hpp
```diff
@@ -174,6 +174,22 @@ class LMHSMessage : public IRIMessage {
   std::string _latestSolidMilestoneHash;
 };
 
+class LMSIMessage : public IRIMessage {
+ public:
```
th0br0, line 178: you might want to delete the default constructor?

tanglescope/common/iri.hpp
```diff
@@ -174,6 +174,22 @@ class LMHSMessage : public IRIMessage {
...
+  explicit LMSIMessage(std::string_view from) {
+    auto chunks = IRIMessage::chunks(from);
+    this->_latestSolidMilestoneIndex = std::stoull(std::move(chunks[1]));
+  }
+
+  inline IRIMessageType type() { return IRIMessageType::LMSI; }
+
+ public:
+  const uint64_t& latestSolidMilestoneIndex() { return this->_latestSolidMilestoneIndex; }
```
th0br0: shouldn't return by reference here. a reference is just as large as the actual value anyway.

tanglescope/statscollector/statscollector.cpp
```diff
@@ -115,11 +117,13 @@ void ZMQCollectorImpl::collect(uint32_t bundleConfirmationHistogramRange,
       rxcpp::observable<>::create<std::shared_ptr<iri::IRIMessage>>(
           [&](auto s) { zmqPublisher(std::move(s), _zmqURL); });
 
+  uint64_t lmsi = 0;
```
th0br0: add a comment to explain what `lmsi` stands for?

th0br0 (changes requested):

tanglescope/common/iri.hpp
tsvisabo: was overlooking, thanks

tanglescope/statscollector/statscollector.cpp
tsvisabo: yup

tanglescope/common/iri.hpp
tsvisabo: Defining a one parameter constructor will prevent the compiler from generating default one,
so no unintentional usage of that is possible

th0br0 (approved):

PR #102, opened 2018-07-31
Fix missing template arguments before '(' token

howjmay:
Error Messege is showed in the following
===================Error Messege==================
cppclient/api_json.cc:275:46:
error: missing template arguments before '(' token
         confirmedBundlesMap.emplace(std::pair(address, b));


th0br0 (approved):

th0br0:
Please use a modern compiler. We use GCC8 / clang6 which do not require the explicit arguments here.
However, I've merged this for now.

PR #120, opened 2018-08-07
cppclient: add unit test of getNodeInfo()

howjmay:
developed with gcc 5.5.0

Imitate the unit test of getBalances(), and there is another unit test way which is imitated from iota.lib.cpp.


th0br0:
Hey @HowJMay ! 
Thanks for the PR. Could you clean up the commits so it only contains your relevant changes?

howjmay:
ok~

jserv:
@HowJMay, You can use `git rebase` to rework these commits.

PR #125, opened 2018-08-08
cppclient: add unit test of getNodeInfo()

howjmay:
Imitate the unit test of getBalances(), and there is another unit test way which is imitated from iota.lib.cpp.

### DIFF
--- a/cppclient/tests/api_json.cc
+++ b/cppclient/tests/api_json.cc
@@ -63,4 +63,30 @@ TEST_F(IotaJsonAPITest, GetBalances) {
   EXPECT_EQ(response, expected);
 }
 
+TEST_F(IotaJsonAPITest, GetNodeInfo) {
+  MockAPI api;
+
+  json req;
+  req["command"] = "getNodeInfo";
+  json res;
+  res["latestMilestone"] = "VBVEUQYE99LFWHDZRFKTGFHYGDFEAMAEBGUBTTJRFKHCFBRTXFAJQ9XIUEZQCJOQTZNOOHKUQIKOY9999";
+  res["latestMilestoneIndex"] = 107;
+  res["latestSolidSubtangleMilestoneIndex"] = 107;
+  
+  cppclient::NodeInfo expected = {res["latestMilestone"], res["latestMilestoneIndex"]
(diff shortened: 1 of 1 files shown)


howjmay:
@jserv  I fixed it

cppclient/tests/api_json.cc
```diff
@@ -62,5 +62,38 @@ TEST_F(IotaJsonAPITest, GetBalances) {
   EXPECT_EQ(api.request, req);
   EXPECT_EQ(response, expected);
 }
+TEST_F(IotaJsonAPITest, GetNodeInfo) {
```
jserv: There should be one blank line to split `GetNodeInfo` item and the previous one.

cppclient/tests/api_json.cc
```diff
@@ -62,5 +62,38 @@ TEST_F(IotaJsonAPITest, GetBalances) {
...
+  MockAPI api;
+
+  json req;
+  req["command"] = "getNodeInfo";
+  json res;
+  char *str = "VBVEUQYE99LFWHDZRFKTGFHYGDFEAMAEBGUBTTJRFKHCFBRTXFAJQ9XIUEZQCJOQTZNOOHKUQIKOY9999";
```
jserv: Avoid C-style string. Instead, make use of C++ `std::string`, which provides abstractions along with better memory management.

cppclient/tests/api_json.cc
```diff
@@ -62,5 +62,38 @@ TEST_F(IotaJsonAPITest, GetBalances) {
...
+  auto response = api.getNodeInfo();
+  
+  // Imitae the unit test in getBlances in entangled
+  EXPECT_EQ(api.request, req);
+  EXPECT_STREQ((*response).latestMilestone.c_str(), str);
+  EXPECT_GE((*response).latestMilestoneIndex, res["latestMilestoneIndex"]);
+  EXPECT_GE((*response).latestSolidMilestoneIndex, res["latestSolidSubtangleMilestoneIndex"]);
+
+
+  /*
```
jserv: Don't leave the following dead code.

Issue #77, opened 2018-07-23
Logger facility


thibault-martinez:
Currently, two logging facilities are being used in different branches.

- https://github.com/embear/logger that turned out not being thread-safe so locks were added on top of it
- A thread-safe logger facility that was made from scratch

This issue aims at merging the best out of the two current logger facilities

tsvisabo:
Maybe @th0br0  can be the judge of it ?

tsvisabo:
@thibault-martinez 
I think we should go with `embear`'s for now since it has been there for a while and is fully tested +
it has contexts support (`logger_id_t`)
can register multiple files (including stdout/stderr) to different error levels, and is highly configurable including prefixing each log with date/line/function etc.. and colors..

thibault-martinez:
@tsvisabo Totally agree

thibault-martinez:
@tsvisabo I would add macros on top of that though.

Something like

```
#define log_trace(...) logger_log(LOG_TRACE, __VA_ARGS__)
#define log_debug(...) logger_log(LOG_DEBUG, __VA_ARGS__)
#define log_info(...) logger_log(LOG_INFO, __VA_ARGS__)
#define log_warn(...) logger_log(LOG_WARN, __VA_ARGS__)
#define log_error(...) logger_log(LOG_ERROR, __VA_ARGS__)
#define log_fatal(...) logger_log(LOG_FATAL, __VA_ARGS__)
```

I find the name and need to specify the level quite verbose as we are going to abuse it.

`logger_helper_print(CONNECTION_LOGGER_ID, LOGGER_ERR, "Failed in creating index, statement: %s", statement);`

What do you think ?

tsvisabo:
@thibault-martinez 
sounds good

thibault-martinez:
Closed by #127

PR #127, opened 2018-08-08
Logger facility

thibault-martinez:
**MUST REVIEW https://github.com/iotaledger/rules_iota/pull/11 BEFORE**

Address #77 

- Removes the old logger
- Add macros on top of the new logger
- Moves the new logger to `/utils`
- Add/edit some logs

# Test Plan:
Not applicable


common/network/components/broadcaster.c
```diff
@@ -65,30 +72,41 @@ bool broadcaster_on_next(broadcaster_state_t *const state,
...
       CONCURRENT_QUEUE_SUCCESS) {
+    log_warning(broadcaster_component_logger,
+                "Pushing to broadcaster queue failed\n");
     return false;
   }
   return true;
 }
 
 bool broadcaster_stop(broadcaster_state_t *const state) {
+  bool ret = true;
```
tsvisabo, line 83: functions should return `retcode_t`

common/network/components/receiver.c
```diff
@@ -9,11 +9,14 @@
 #include "ciri/node.h"
 #include "common/logger_helper.h"
 
+static char const receiver_component_logger[] = "receiver_component";
```
tsvisabo: Any reason to prefer that over #define, there are already some components (storage) with defines, we should align to the same convention

tsvisabo (changes requested):

common/network/components/broadcaster.c
thibault-martinez: Agreed, I'll open an issue to handle that separately

thibault-martinez: https://github.com/iotaledger/entangled/issues/129

common/network/components/receiver.c
thibault-martinez: Generally I tend to use static const vars instead of #define as they provide type-safety and debugging symbols. But there is a memory overhead so maybe in this context #define are perfectly fine.

tsvisabo (approved):

PR #134, opened 2018-08-10
Requester component & core structure

thibault-martinez:
This PR implements the requester component.

Notable changes and/or additions:

- Adds the requester component
- Fixes some storage compilation warnings
- Adds a `core` (name to be discussed) structure that bind everything together (database, node, api, ixi...)
- Adds a database to cIRI

# Test Plan:
WIP


thibault-martinez:
Blocked by #135

ciri/core.c
```diff
@@ -0,0 +1,86 @@
...
+// TODO(thibault) from configuration variables
+#define CIRI_DB_PATH "ciri/ciri.db"
+#define CORE_LOGGER_ID "core"
+
+bool core_init(core_t* const core) {
+  if (core == NULL) {
+    return false;
+  }
+  logger_helper_init(CORE_LOGGER_ID, LOGGER_DEBUG, true);
+  memset(core, 0, sizeof(core_t));
```
th0br0: could we move this below the whiteline?

ciri/core.c
```diff
@@ -0,0 +1,86 @@
...
+  log_info(CORE_LOGGER_ID, "Initializing database connection\n");
+  core->db_conf.db_path = CIRI_DB_PATH;
+  core->db_conf.index_address = true;
+  core->db_conf.index_approvee = true;
+  core->db_conf.index_bundle = true;
+  core->db_conf.index_tag = true;
+  if (init_connection(&core->db_conn, &core->db_conf) != RC_OK) {
+    log_critical(CORE_LOGGER_ID, "Initializing database connection failed\n");
+    return false;
+  }
```
th0br0: insert a whiteline?

ciri/core.c
```diff
@@ -0,0 +1,86 @@
...
+  log_info(CORE_LOGGER_ID, "Initializing cIRI node\n");
+  if (node_init(&core->node, core) == false) {
```
th0br0: could this also return `RC_OK` or similar?

ciri/core.c
```diff
@@ -0,0 +1,86 @@
...
+  }
+  return true;
+}
+
+bool core_start(core_t* const core) {
+  if (core == NULL) {
+    return false;
+  }
+  log_info(CORE_LOGGER_ID, "Starting cIRI node\n");
+  if (node_start(&core->node) == false) {
```
th0br0: `node_start` should return a return code?

common/network/components/requester.c
```diff
@@ -6,52 +6,122 @@
...
+  if (state == NULL) {
+    return false;
+  }
+  if (node == NULL) {
     return false;
   }
   logger_helper_init(REQUESTER_COMPONENT_LOGGER_ID, LOGGER_DEBUG, true);
-  if (INIT_CONCURRENT_QUEUE_OF(trit_array_p, state->queue) !=
-      CONCURRENT_QUEUE_SUCCESS) {
+  if (INIT_CONCURRENT_LIST_OF(trit_array_p, state->list, trit_array_cmp) !=
```
th0br0, line 26: why no queue?
it's a priority queue on IRI I think? or is that just for the processor

common/network/components/requester.c
```diff
@@ -6,52 +6,122 @@
...
+                  "Transactions requests list is full\n");
+    }
+  }
   return true;
 }
 
-trit_array_p get_transaction_to_request(requester_state_t *const state) {
-  trit_array_p hash;
+bool get_transaction_to_request(requester_state_t *const state,
+                                trit_array_p *hash) {
```
th0br0, line 99: trit_array_p already is a pointer

th0br0 (changes requested):

common/network/components/requester.c
thibault-martinez: IRI has a `LinkedHashSet` for that.
The queue API doesn't allow us to do what's needed here (contains/remove).
We'll need to replace the list with an iterable hashmap.

common/network/components/requester.c
thibault-martinez: Yes, but we want to modify the pointer, because the list stores `trit_array_p`

ciri/core.c
```diff
@@ -0,0 +1,93 @@
...
+  if (node_stop(&core->node)) {
+    log_error(CORE_LOGGER_ID, "Stopping cIRI node failed\n");
+    ret = RC_CORE_FAILED_NODE_STOP;
+  }
+  return ret;
+}
+
+retcode_t core_destroy(core_t* const core) {
+  retcode_t ret = RC_OK;
+
```
th0br0, line 82: How do you guarantee that core is stopped before it's destroyed? I guess that might also apply to other components

th0br0 (changes requested):

ciri/core.c
thibault-martinez: As discussed, the field `running`is now checked before destroying a component.
It has also been added to `node` and `core`.

ciri/node.c
```diff
@@ -90,26 +107,33 @@ retcode_t node_stop(node_t* const node) {
...
     ret = RC_NODE_FAILED_RECEIVER_STOP;
   }
+
   log_info(NODE_LOGGER_ID, "Stopping responder component\n");
   if (responder_stop(&node->responder) == false) {
     log_error(NODE_LOGGER_ID, "Stopping responder component failed\n");
     ret = RC_NODE_FAILED_RESPONDER_STOP;
   }
+
+  node->running = false;
```
th0br0: You probably want to set this before entering the stop routines to avoid possible races?

th0br0 (approved):

ciri/node.c
thibault-martinez: If I set it before, we can actually destroy the node with the components still running ?

Issue #138, opened 2018-08-13
Add tests that will cover storage functionality completely


tsvisabo:
Tests for multi threaded access
Tests for multiple rows that matches a condition

PR #139, opened 2018-08-13
Storage - Load hashes only

tsvisabo:
# Test Plan:
One test is Included
Opened issue #138

### DIFF
--- a/common/storage/sql/sqlite3/storage.c
+++ b/common/storage/sql/sqlite3/storage.c
@@ -26,6 +26,15 @@ typedef struct iota_transactions_pack {
   size_t num_loaded;
 } iota_transactions_pack;
 
+typedef struct iota_hashes_pack {
+  trit_array_p* hahses;
+  size_t num_hashes;
+  size_t num_loaded;
+} iota_hashes_pack;
+
+int select_hashes_cb(void* hashes, int num_cols, char** col_values,
+                     char** col_names);
+
 retcode_t iota_stor_init() {
   logger_helper_init(SQLITE3_LOGGER_ID, LOGGER_INFO, true);
 }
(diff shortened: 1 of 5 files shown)


common/storage/sql/sqlite3/storage.c
```diff
@@ -26,6 +26,15 @@ typedef struct iota_transactions_pack {
   size_t num_loaded;
 } iota_transactions_pack;
 
+typedef struct iota_hashes_pack {
+  trit_array_p* hahses;
```
th0br0: typo

common/storage/sql/sqlite3/storage.c
```diff
@@ -140,6 +149,50 @@ retcode_t iota_stor_exist(const connection_t* const conn,
...
+                                       trit_array_p data_out[],
+                                       size_t max_num_hashes,
+                                       size_t* num_loaded) {
+  char statement[MAX_SELECT_STATEMENT_SIZE];
+  iota_transactions_select_hashes_statement(col_name, key, statement,
+                                            MAX_SELECT_STATEMENT_SIZE);
+
+  char* err_msg = 0;
+  iota_hashes_pack pack;
+  pack.hahses = data_out;
```
th0br0: typo

common/storage/sql/sqlite3/storage.c
```diff
@@ -140,6 +149,50 @@ retcode_t iota_stor_exist(const connection_t* const conn,
...
+                                       const trit_array_p key,
+                                       trit_array_p data_out[],
+                                       size_t max_num_hashes,
+                                       size_t* num_loaded) {
+  char statement[MAX_SELECT_STATEMENT_SIZE];
+  iota_transactions_select_hashes_statement(col_name, key, statement,
+                                            MAX_SELECT_STATEMENT_SIZE);
+
+  char* err_msg = 0;
+  iota_hashes_pack pack;
```
th0br0: this should be passed into the method?

common/storage/sql/statements.c
```diff
@@ -127,3 +127,27 @@ retcode_t iota_transactions_update_statement(const char *index_col,
                                              const iota_transaction_t tx,
                                              char statement[],
                                              size_t statement_cap) {}
+
+retcode_t iota_transactions_select_hashes_statement(const char *index_col,
+                                                    const trit_array_p key,
```
th0br0, line 132: Let's use an `enum` for this kind of key?

common/storage/sql/statements.c
```diff
@@ -127,3 +127,27 @@ retcode_t iota_transactions_update_statement(const char *index_col,
...
+    memcpy(key_str, key->trits, key->num_bytes);
+    key_str[key->num_bytes] = 0;
+    res = snprintf(statement, statement_cap, "SELECT %s FROM %s WHERE %s='%s'",
+                   COL_HASH, TRANSACTION_TABLE_NAME, index_col, key_str);
+  }
+
+  if (res < 0 || res == statement_cap) {
+    log_error(SQL_STATEMENTS_ID, "Failed in creating statement, statement: %s",
+              statement);
+    return RC_SQL_FAILED_WRITE_STATEMENT;
```
th0br0, line 151: this is a `read` statement, not a `write` statement

th0br0 (changes requested):

common/storage/sql/sqlite3/tests/test_sqlite3.c
```diff
@@ -73,6 +74,28 @@ void test_stored_transaction(void) {
   }
 }
 
+void test_stored_load_hashes_by_address(void) {
+  trit_array_p hashes[5];
+
+  for (int i = 0; i < 5; ++i) {
+    hashes[i] = trit_array_new(FLEX_TRIT_SIZE_243);
```
thibault-martinez: Shouldn't it be just 243 here (not flex) ?
`trit_array_new` already takes care of the conversion.

common/storage/sql/sqlite3/storage.c
tsvisabo: yes, that would be better

common/storage/sql/statements.c
tsvisabo: Yes, but we are writing the read statement :)

common/storage/sql/sqlite3/tests/test_sqlite3.c
tsvisabo: That is correct, thanks

common/storage/sql/statements.c
tsvisabo: But the encoding can be different, so the enumToKeyTable (if that is what you have in mind) will need entry per column per encoding (1/3/4/5)

th0br0: Nevermind, I was confusing index_col & key

th0br0 (approved):

Issue #59, opened 2018-07-16
Design configuration variables management


thibault-martinez:
- [x] Command line arguments (#146)
- [ ] Configuration file

PR #146, opened 2018-08-14
Initialize TCP/UDP receiver services with ports from configuration variables

### LINKED ISSUE #104: Initialize TCP/UDP receiver services with ports from configuration variables
For now, receiver component is initialized with hardcoded values for its TCP and UDP receiver services.

https://github.com/iotaledger/entangled/pull/84/files/b8cf71abb68444bd562d4ba29ed099ba40304f91#diff-efa82e9a710e5a89a3471e8d47865cc8L10

jkrvivian:
Fixes #104 

* Configuration variables can be declared in `ciri/conf/configuration.h`
* A simple argument parser implemented with `getopt_long` can change the default value in `ciri/conf/configuration.c`, there are 2 arguments now:
    * `[--udp-receiver-port | -u] [port_number]`
    * `[--tcp-receiver-port | -t] [port_number]`
* Arguments are declared in `ciri/conf/config_args.h`

## Test Plan:
After bazel build, run `$ bazel-bin/ciri/ciri [argument]`, the log will show the port you defined.


th0br0:
Thanks for the PR!

Unfortunately, getopt is not available on all platforms. We should abstract here.

jserv:
The availability of `getopt()` is required for POSIX.1-2001. @th0br0, did you mean entangled needs something like [pg_getopt.h](https://github.com/postgres/postgres/blob/master/src/include/pg_getopt.h) as PostgreSQL does?

Reference: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getopt.html

jkrvivian:
I chose to use `getopt_long` because it supported both short and long options(same with IRI now), but I didn't notice it's a GNU extension and the availability on other platforms.

@jserv , thanks for giving a great example!
I'll implement our own if `getopt_long` is not available just like PostgreSQL does.

@th0br0 how do you think about it?
And I notice this PR is also related to #59 :thinking:

th0br0:
> did you mean entangled needs something like pg_getopt.h as PostgreSQL does?
We definitely need some indirection (i.e. an intermediate configuration struct or similar) to hold the actual settings. This way, whether we use getopt or sth else becomes irrelevant. 
(Don't use globals!)

Given that we're mostly testing on Linux/OSX right now, we can use `getopt` without worrying too much about it.

So:
Something like
`ciri/conf/config.h`
```
typedef struct {
  uint16_t neighbour_port_tcp;
  uint16_t neighbour_port_udp;
// ...
} ciri_config_t;
```
`ciri/conf/provider.h`
```
retcode_t ciri_conf_parse(ciri_config_t*, size_t argc, char** argv);
```

And then `getopt` just being an OS-specific implementation which we link against by default

`ciri/conf/getopt.c`
```
retcode_t ciri_conf_parse(ciri_config_t* out, size_t argc, char** argv) {
  getopt_long //...
```

Sounds like the most sane & portable approach here
@thibault-martinez what's your take?

ciri/conf/config.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __COMMON_CONFIGURATION__
```
thibault-martinez: Please follow the scoping pattern for include guards.
`__CIRI_CONF_CONFIG_H__` for example.

ciri/conf/config.h
```diff
@@ -0,0 +1,18 @@
...
+#define __COMMON_CONFIGURATION__
+#include <stdint.h>
+
+typedef struct {
+    uint16_t tcp_port_g;
```
thibault-martinez: `_g` was for global, you can remove them now.

ciri/conf/config.h
```diff
@@ -0,0 +1,18 @@
...
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __COMMON_CONFIGURATION__
+#define __COMMON_CONFIGURATION__
+#include <stdint.h>
+
+typedef struct {
```
thibault-martinez: `typedef struct ciri_config_s` so that it can be forward declared later.

ciri/conf/config_args.h
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __COMMON_CONFIGURATION_ARGS__
```
thibault-martinez: scoping pattern

ciri/conf/config_args.h
```diff
@@ -0,0 +1,17 @@
...
+#define __COMMON_CONFIGURATION_ARGS__
+
+char* short_options = "u:t:";
+struct option long_options[] = {
```
thibault-martinez: This structure probably won't be available if `getopt_long`itself is not.

ciri/conf/getopt.c
```diff
@@ -0,0 +1,34 @@
...
+          out->tcp_port_g = atoi(optarg);
+          break;
+        case 'u':
+          out->udp_port_g = atoi(optarg);
+          break;
+        default:
+          break;
+      }
+  }
+  return 0;
```
thibault-martinez: `return RC_OK`

ciri/conf/provider.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __COMMON_CONFIG_PROVIDER__
```
thibault-martinez: scoping pattern

ciri/conf/provider.h
```diff
@@ -0,0 +1,18 @@
...
+#define __COMMON_CONFIG_PROVIDER__
+#include <getopt.h>
```
thibault-martinez: insert newline please

ciri/main.c
```diff
@@ -9,13 +9,15 @@
 
 #include "ciri/core.h"
 #include "ciri/node.h"
+#include "ciri/conf/provider.h"
 #include "utils/logger_helper.h"
 
 #define MAIN_LOGGER_ID "main"
 
 static core_t core_g;
+static ciri_config_t* config;
```
thibault-martinez: Just store it in the core so that we don't have to pass it everywhere

thibault-martinez (changes requested):
Hey @jkrvivian, thanks for the PR !
Just some quick changes before I dive into it.

ciri/main.c
jkrvivian: Agree, but we can't get `config` if we wrapped it in core in `node.c`, since we forward declared core here.

We'll need it in `bool receiver_init(receiver_state_t *const state, node_t *const node, uint16_t tcp_port, uint16_t udp_port);`

ciri/conf/config_args.h
jkrvivian: Do you mean if `getopt_long` is not available?
Then, we need to do some checking like @jserv mentioned what PostgreSQL did, right?

ciri/main.c
thibault-martinez: I'm not sure to understand what you mean. Node has access to the core, then node will have access to config.

ciri/conf/getopt.c
```diff
@@ -0,0 +1,34 @@
+#include <stdlib.h>
+#include <string.h>
+
+#include "ciri/conf/provider.h"
+#include "ciri/conf/config_args.h"
+
+ciri_config_t* ciri_conf_init() {
+  ciri_config_t *config = (ciri_config_t* )malloc(sizeof(ciri_config_t));
```
thibault-martinez: If we put an instance of `ciri_config_t` in the core, we can get rid of this allocation, by the way it was leaking.

ciri/conf/getopt.c
```diff
@@ -0,0 +1,31 @@
+#include <stdlib.h>
+
+#include "ciri/conf/provider.h"
+#include "ciri/conf/config_args.h"
+
+retcode_t ciri_conf_init(ciri_config_t* config) {
+  if (config == NULL) {
+      return 1;
```
jkrvivian: Should I just add a return code like `RC_CORE_FAILED_CONF_INIT` here?

thibault-martinez: Yes, please add a new CONF module in common/errors.h on the same pattern than the others.

ciri/conf/config.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __CIRI_CONF_CONFIG_H__
+#define __CIRI_CONF_CONFIG_H__
+#include <stdint.h>
```
thibault-martinez: Insert newline please

ciri/conf/getopt.c
```diff
@@ -0,0 +1,31 @@
...
+  }
+  config->tcp_port = 14260;
+  config->udp_port = 14261;
+
+  return RC_OK;
+}
+
+retcode_t ciri_conf_parse(ciri_config_t* out, size_t argc, char** argv) {
+  int arg;
+  while ((arg = getopt_long(argc, argv, short_options, long_options, NULL)) != -1) {
```
thibault-martinez: Insert new line

ciri/conf/getopt.c
```diff
@@ -0,0 +1,31 @@
...
+
+#include "ciri/conf/provider.h"
+#include "ciri/conf/config_args.h"
+
+retcode_t ciri_conf_init(ciri_config_t* config) {
+  if (config == NULL) {
+      return 1;
+  }
+  config->tcp_port = 14260;
+  config->udp_port = 14261;
```
thibault-martinez: Default values shouldn't be in providers.
You can #define them outside and use them here.

ciri/conf/provider.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __CIRI_CONF_PROVIDER_H__
+#define __CIRI_CONF_PROVIDER_H__
+#include <getopt.h>
```
thibault-martinez: insert newline

ciri/core.c
```diff
@@ -38,7 +37,7 @@ retcode_t core_init(core_t* const core) {
   }
 
   log_info(CORE_LOGGER_ID, "Initializing cIRI node\n");
-  if (node_init(&core->node, core)) {
+  if (node_init(&core->node, core, &core->config)) {
```
thibault-martinez: You don't even need to pass config since you have core

ciri/main.c
```diff
@@ -25,6 +27,16 @@ int main() {
   logger_output_register(stdout);
   logger_output_level_set(stdout, LOGGER_DEBUG);
   logger_helper_init(MAIN_LOGGER_ID, LOGGER_DEBUG, true);
+  memset(&core_g, 0, sizeof(core_t));
+
+  // configuration argument parser
+  if (ciri_conf_init(&core_g.config)) {
+      return EXIT_FAILURE;
```
thibault-martinez: add log

ciri/main.c
```diff
@@ -25,6 +27,16 @@ int main() {
...
+  }
+
+  if (ciri_conf_parse(&core_g.config, argc, argv)) {
+      return EXIT_FAILURE;
```
thibault-martinez: add log

thibault-martinez (changes requested):

ciri/core.c
jkrvivian: In `node.h`, we forward declare `core_t`, so we can't access `config` or other variables of `core` in `node.c`.
But if we include `core.h` in `node.c` or `node.h`, it will cause a dependency cycle when bazel build.

thibault-martinez: Just `#include "ciri/core.h"`in `node.c`, you shouldn't have any problem

thibault-martinez: At least, I don't

ciri/conf/getopt.c
jkrvivian: How about #define in `config_args.h`?

thibault-martinez: I would prefer another file with just default values

jkrvivian: Sure!

ciri/conf/provider.h
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __CIRI_CONF_PROVIDER_H__
+#define __CIRI_CONF_PROVIDER_H__
+
+#include <getopt.h>
```
thibault-martinez: This should be in `ciri/conf/getopt.c` as `ciri/conf/provider.h` is supposed to be abstract

thibault-martinez (changes requested):

thibault-martinez:
Hi @jkrvivian, if you plan to do other PRs, please be sure to run `./hooks/autohook.sh install` at some point ! It will install the hooks, including style checking.

jkrvivian:
@thibault-martinez , there're more modules added to return code in the previous commit.
Since the `config` is in `core` now, maybe we should also move the `config` return code to `core` module.
What do you think?

ciri/node.h
```diff
@@ -14,6 +14,7 @@
 #include "common/network/components/receiver.h"
 #include "common/network/components/requester.h"
 #include "common/network/components/responder.h"
+#include "ciri/conf/config.h"
```
thibault-martinez: Not needed

thibault-martinez (changes requested):

thibault-martinez:
Hi @jkrvivian , I'm fine with letting config errors in their own module.
Please address the requested changes, solve conflicts and we'll be good !

thibault-martinez (approved):

thibault-martinez:
Thanks @jkrvivian !

jkrvivian:
@thibault-martinez , thanks for your review and advices! :laughing:

PR #148, opened 2018-08-14
common: make trit_encoding a build parameter

th0br0:
This allows the user to specify the internal trit-memory encoding that entangled should use at build time.

Defaults to trytes if not specified.
Once merged, we'll add build targets for each encoding type to CI

# Test Plan:
Tests run.

### DIFF
--- a/common/model/BUILD
+++ b/common/model/BUILD
@@ -12,78 +12,12 @@ cc_library(
 )
 
 cc_library(
-    name = "transaction_3",
-    srcs = ["transaction.c"],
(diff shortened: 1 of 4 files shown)


thibault-martinez:
Cool, how do we use that from the CLI ?

thibault-martinez (approved):

th0br0:
`--define trit_encoding=5`, see https://docs.bazel.build/versions/master/configurable-attributes.html
note that storage currently depends on trit_encoding=3

PR #165, opened 2018-08-19
Fix compilation warnings

thibault-martinez:
This PR fixes some compilation warnings.

# Test Plan:
Not applicable.

### DIFF
--- a/utils/files.c
+++ b/utils/files.c
@@ -5,6 +5,12 @@
  * Refer to the LICENSE file for licensing information
  */
 
+#ifdef _WIN32
+#include <io.h>
+#else
+#include <unistd.h>
+#endif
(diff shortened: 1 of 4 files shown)


utils/files.c
```diff
@@ -5,6 +5,12 @@
  * Refer to the LICENSE file for licensing information
  */
 
+#ifdef _WIN32
```
th0br0: could you do a check similar to the one in the threading utils there?

th0br0 (changes requested):

utils/files.c
thibault-martinez: Sure

th0br0 (approved):

PR #175, opened 2018-08-21
cclient: implement iota_api_get_node_info

oopsmonk:
This PR implements iota_api_get_node_info.

# Test Plan:
Unit tests in //cclient/serialization:test_json

### DIFF
--- a/cclient/serialization/json/json_serializer.c
+++ b/cclient/serialization/json/json_serializer.c
@@ -40,6 +40,52 @@ retcode_t utarray_to_json_array(UT_array* ut, cJSON* json_root,
   return RC_OK;
 }
 
+retcode_t json_get_size(const cJSON* json_obj, const char* obj_name,
+                        size_t* num) {
+  cJSON* json_value = cJSON_GetObjectItemCaseSensitive(json_obj, obj_name);
+  if (json_value == NULL) return RC_CCLIENT_JSON_KEY;
+
+  if (cJSON_IsNumber(json_value)) {
+    *num = (size_t)json_value->valuedouble;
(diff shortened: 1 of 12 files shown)


cclient/serialization/json/json_serializer.c
```diff
@@ -40,6 +40,52 @@ retcode_t utarray_to_json_array(UT_array* ut, cJSON* json_root,
   return RC_OK;
 }
 
+retcode_t json_get_size(const cJSON* json_obj, const char* obj_name,
+                        size_t* num) {
+  cJSON* json_value = cJSON_GetObjectItemCaseSensitive(json_obj, obj_name);
+  if (json_value == NULL) return RC_CCLIENT_JSON_KEY;
+
+  if (cJSON_IsNumber(json_value)) {
```
tsvisabo, line 48: Can it be negative? would you return error in that case?

tsvisabo (changes requested):

cclient/serialization/json/json_serializer.c
oopsmonk: it's a boolean and return RC_CCLIENT_JSON_PARSE if the value is not a number.

tsvisabo: Yes, i was referring to: `(size_t)json_value->valuedouble`
could that line be casting negative number to `size_t`?

oopsmonk: could the value be negative in the response? ex. memory size, time, index of a milestone...

tsvisabo (approved):

PR #209, opened 2018-08-29
Correctly protect the freed variables from access again

furuame:
These assignments are redundant because of `pass-by-value` in C.

# Test Plan:
CI must pass.


oopsmonk:
Could you explain more in your perspective?
It's a safety for the structure. it prevents to access the member of the structure after freeing.

furuame:
The variable what you assigned `NULL` is the copy of the pointer. It cannot prevent the member of the structure to be accessed safely "outside" the function after freed. The pointer to the pointer should be passed to achieve your original goal.

```
// In the function "find_transactions_req_free"
(gdb) p req
$2 = (find_transactions_req_t *) 0x0

// Outside the function "find_transactions_req_free"
(gdb) p find_tran
$3 = (find_transactions_req_t *) 0x55555556be70
```

oopsmonk:
You're right, thanks.

thibault-martinez (approved):

PR #265, opened 2018-09-09
API base

thibault-martinez:
This PR adds a basic API struct and related functions.
API is also launched by cIRI.

# Test Plan:
Command line


tsvisabo (approved):

th0br0 (commented):
Does it really make sense to tie a threading model to the actual API design? 
Also, this specifies a port. What about APIs that are not bound to a specific port but are exposed via IPC or similar means?

PR #270, opened 2018-09-11
cclient: added extended API header file.

oopsmonk:
This PR reviews extended API design in cclient, those extended API functions refer to the [iota/core](https://github.com/iotaledger/iota.js/tree/next/packages/core) in `iota.js`.
`TODO` objects and object operations should be designed and implemented during the API implementation.

# Test Plan:
No

### DIFF
--- a/cclient/iota_extended_api.h
+++ b/cclient/iota_extended_api.h
@@ -32,7 +32,7 @@ typedef struct {
 
 typedef struct {
   int64_t balance;
-  int32_t keyIndex;
+  int32_t key_index;
   int security;
   hash_t address;
 } input_t;
--- a/cclient/iota_extended_api.h
+++ b/cclient/iota_extended_api.h
@@ -1,0 +1,427 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef CCLIENT_IOTA_EXTENDED_API_H_
+#define CCLIENT_IOTA_EXTENDED_API_H_
+
+#include "iota_api.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+//----TODO objects----------
+typedef char_buffer_t hash_t;
+typedef UT_array hashes_t;
+typedef UT_array address_list_t;
+typedef UT_array inputs_list_t;
+typedef UT_array transfer_list_t;
+typedef UT_array transaction_list_t;
+typedef UT_array transaction_objs_t;
+typedef UT_array booleans_t;
+typedef UT_array input_list_t;
+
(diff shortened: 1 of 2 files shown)


cclient/iota_extended_api.h
```diff
@@ -0,0 +1,427 @@
...
+typedef UT_array input_list_t;
+
+typedef struct {
+  iota_http_service_t http;
+  serializer_t serializer;
+} iota_service_t;
+
+typedef struct {
+  int64_t balance;
+  int32_t keyIndex;
```
tsvisabo: snake_case

tsvisabo (changes requested):

cclient/iota_extended_api.h
```diff
@@ -0,0 +1,427 @@
...
+extern "C" {
+#endif
+
+//----TODO objects----------
+typedef char_buffer_t hash_t;
+typedef UT_array hashes_t;
+typedef UT_array address_list_t;
+typedef UT_array inputs_list_t;
+typedef UT_array transfer_list_t;
+typedef UT_array transaction_list_t;
```
tsvisabo: that definitions are useful for other cases, can you have them in `common`?

tsvisabo (changes requested):

cclient/iota_extended_api.h
oopsmonk: it's in the plan, most of those objects should be in the `common`.  :wink:

thibault-martinez:
A general comment about the naming of these functions.
Maybe we want to reserve the `iota_api_*` scope for the actual cIRI API ?
What about `iota_client_*` ?

oopsmonk:
> A general comment about the naming of these functions.
> Maybe we want to reserve the `iota_api_*` scope for the actual cIRI API ?
> What about `iota_client_*` ?

`iota_client_*` sounds good for cclient, but how about  `iota_iri_*` for cIRI since `com.iota.iri.*` packages in IRI?

cclient/iota_extended_api.h
```diff
@@ -0,0 +1,429 @@
...
+  transfer_list_t transfers;
+  transaction_list_t transactions;
+} account_data_t;
+
+typedef struct {
+  int security;
+  int32_t start;
+  int32_t end;
+} address_opt_t;
+//---------END of TODO objects-------------
```
th0br0: Can we get rid of the TODO here?

th0br0 (approved):

th0br0:
We could also just just not have a `iota_` prefix? :thinking:

thibault-martinez:
We definitely need to have a naming convention discussion :)

oopsmonk:
> We could also just just not have a `iota_` prefix? thinking

do you mean both of file name and API functions?

tsvisabo (approved):

PR #275, opened 2018-09-12
Snapshot WIP

### LINKED ISSUE #280: Implement snapshot

thibault-martinez:
Fixes #280

# Test Plan:
WIP


common/trinary/tryte_ascii.c
```diff
@@ -0,0 +1,26 @@
...
+#include "common/trinary/tryte_ascii.h"
+
+#define TRYTE_ALPHABET "9ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
+bool ascii_to_trytes(char const *const input, tryte_t *const output) {
+  int j = 0, dec = 0, first = 0, second = 0;
+
+  for (int i = 0; input[i]; i++) {
+    dec = input[i];
+    if (dec > 255) {
```
tsvisabo: can not happen if input is `char*`

common/trinary/tryte_ascii.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <stdbool.h>
+
+#include "common/trinary/tryte.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+bool ascii_to_trytes(char const *const input, tryte_t *const output);
```
tsvisabo: should we not have the reverse function as well?

consensus/snapshot/snapshot.c
```diff
@@ -10,12 +10,117 @@
...
+      goto done;
+    }
+    *delim = '\0';
+    value = atoll(delim + 1);
+    if (value > 0) {
+      if ((entry = malloc(sizeof(state_entry_t))) == NULL) {
+        ret = RC_SNAPSHOT_OOM;
+        goto done;
+      }
+      memcpy(entry->hash, line, 81);  // TODO
```
tsvisabo: that's in case we expect tryte formatted snapshot file

consensus/snapshot/snapshot.c
```diff
@@ -143,3 +143,11 @@ bool iota_snapshot_is_state_consistent(state_map_t *const state) {
   }
   return true;
 }
+
+size_t iota_snapshot_get_index(snapshot_t *const snapshot) {
+  size_t index;
+  rw_lock_handle_rdlock(&snapshot->rw_lock);
+  index = snapshot->index;
```
tsvisabo: Maybe we should use atomic for that?

consensus/snapshot/snapshot.c
```diff
@@ -148,8 +148,23 @@ bool iota_snapshot_is_state_consistent(state_map_t *const state) {
...
   rw_lock_handle_unlock(&snapshot->rw_lock);
   return index;
 }
+
+int64_t iota_snapshot_get_balance(snapshot_t *const snapshot,
+                                  flex_trit_t *const hash) {
+  state_entry_t *entry;
+  int64_t balance = 0;
+
+  rw_lock_handle_rdlock(&snapshot->rw_lock);
```
tsvisabo: what if address was not found in `state_map`?
let's return `retcode_t`

tsvisabo (changes requested):

common/trinary/tryte_ascii.c
thibault-martinez: Right

common/trinary/tryte_ascii.h
thibault-martinez: Will do

consensus/snapshot/snapshot.c
thibault-martinez: You are reviewing an old commit, this has changed a little but I still expect trytes.
Can it not be trytes ?

consensus/snapshot/snapshot.c
thibault-martinez: Why not, we have an opened issue concerning atomic handles.
In the meantime, this lock has other purposes and it's in reading mode for this.

consensus/snapshot/snapshot.c
tsvisabo: for now only trytes, but i guess we can serialize 4/5 trits per byte in the future
FLEX_TRITS_NUM_243 ?

consensus/snapshot/snapshot.c
thibault-martinez: Sure

consensus/snapshot/snapshot.c
thibault-martinez: Already done in the latest commit ;)

tsvisabo (approved):

tsvisabo (approved):

PR #290, opened 2018-09-16
Use memset_s (or equivalent) when handling sensitive info

rajivshah3:
Changes `memset` to a custom function when clearing sensitive info out of the memory. According to [MSC06-C. Beware of compiler optimizations](https://wiki.sei.cmu.edu/confluence/display/c/MSC06-C.+Beware+of+compiler+optimizations), `memset` could be removed by a compiler during optimization if it does not deem it necessary. `memset_s` and `SecureZeroMemory` are guaranteed to not be removed by the compiler (see https://en.cppreference.com/w/c/string/byte/memset and https://msdn.microsoft.com/en-us/library/windows/desktop/aa366877).

I changed `common/helpers/sign.c` and `mobile/android/Interface.cpp`, but I'm not sure if the `memset`s that are used in other places need to be changed as well.

# Test Plan:

- Existing tests should pass


thibault-martinez:
Some comments we had in private:
- `memset_s` is C11
- The availability of certain C11 features varies from one compiler to another
- Not available on Windows plateforms
- We can create a common entry point that would redirect to the most efficient available solution: `memset_s` if available, `SecureZeroMemory` if windows, custom function otherwise

rajivshah3:
@thibault-martinez I changed this so that it will use whatever is available based on the preprocessor directives

utils/memset_secure.c
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __STDC_LIB_EXT1__
+#define __STDC_LIB_EXT1__
+#endif
```
thibault-martinez: I'm not sure you should do that, can you provide a link where you saw this ?

thibault-martinez (changes requested):

utils/memset_secure.c
rajivshah3: I got it from https://en.cppreference.com/w/c/string/byte/memset, but I realized I misread this part 🤦‍♂️  : 

> As with all bounds-checked functions, memset_s is only guaranteed to be available if `__STDC_LIB_EXT1__` is defined by the implementation and if the user defines `__STDC_WANT_LIB_EXT1__` to the integer constant 1 before including string.h.

Will fix

rajivshah3: Resolved in 0f85d2a

rajivshah3:
Closing in favor of #839

PR #294, opened 2018-09-17
Consensus and common/model: bundle/transfer - shared code

tsvisabo:

### DIFF
--- a/common/model/transfer.c
+++ b/common/model/transfer.c
@@ -10,6 +10,7 @@
 #include "common/helpers/sign.h"
 #include "common/trinary/trit_tryte.h"
 #include "common/trinary/tryte_long.h"
+#include "common/trinary/trit_long.h"
 
 /***********************************************************************************************************
  * Transfer Input data structure
@@ -422,7 +423,6 @@ int transfer_ctx_count(transfer_ctx_t transfer_ctx) {
 int transfer_ctx_hash(transfer_ctx_t transfer_ctx, Kerl *kerl,
                       transfer_t *transfers, size_t len) {
   size_t i, j, count, current_index = 0;
-  tryte_t essence[NUM_TRYTES_ESSENCE];
   trit_t essence_trits[NUM_TRITS_ESSENCE];
 
   // Calculate bundle hash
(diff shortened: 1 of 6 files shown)


common/model/transfer.c
```diff
@@ -667,3 +652,29 @@ void transfer_iterator_free(transfer_iterator_t transfer_iterator) {
...
 }
+
+
+void absorb_essence(Kerl *const kerl, flex_trit_t *address, int64_t value,
+                    flex_trit_t *obsolete_tag, uint64_t timestamp,
+                    int64_t current_index, int64_t last_index,
+                    trit_t *const essence_trits) {
+  memset(essence_trits, 0, NUM_TRITS_ESSENCE);
+
+  flex_trits_to_trits(essence_trits, NUM_TRITS_ADDRESS, address,
```
th0br0: Instead of doing all the additions, can't we have a `trit_t* essence = essence_trits` ptr that we simply move ahead when we write to it? i.e.
```
flex_trits_to_trits(essence, ...);
essence = &essence[NUM_TRITS_ADDRESS];

long_to_trits(value, &essence);
essence = &essence[NUM_TRITS_VALUE];
```

consensus/bundle_validator/bundle_validator.c
```diff
@@ -88,7 +88,8 @@ retcode_t bundle_validate(const tangle_t* const tangle, trit_array_p tail_hash,
 
       *is_valid = true;
       calculate_bundle_hash(bundle, bundle_hash_calculated);
-      if (memcmp(bundle_hash, bundle_hash_calculated, FLEX_TRIT_SIZE_243) != 0) {
+      if (memcmp(bundle_hash, bundle_hash_calculated, FLEX_TRIT_SIZE_243) !=
```
th0br0, line 91: missing some clang-format love?

th0br0 (changes requested):

consensus/bundle_validator/bundle_validator.c
tsvisabo: that's actually clang-formatted, i would have never came up with that

common/model/transfer.c
tsvisabo: done

consensus/bundle_validator/bundle_validator.c
th0br0: Huh, interesting.

th0br0 (approved):

Issue #234, opened 2018-09-05
Setup cIRI database with bazel


thibault-martinez:
Either we ask the user to generate the database as part of the building process (something like `sqlite3 ciri/ciri.db < common/storage/sql/schema.sql`) either bazel does it under the hood.
Issue is that bazel generated files are read-only, so as a temporary solution we pushed a database (`ciri/ciri.db`) containing nothing but the schemas and bazel does a link to it, user has nothing to do but there is a redundancy between the database and the schema (located under `common/storage/sql/schema.sql`)

th0br0:
Why don't we just initialise the database as part of the initial run if it can't be found?

thibault-martinez:
Initialise the database from bazel or from cIRI itself ?

th0br0:
cIRI itself. Woudl need access to the schema I guess.

thibault-martinez:
It has been decided to ask the user to generate the database. See README.

PR #297, opened 2018-09-18
Snapshot signature

### LINKED ISSUE #273: Validate signature of snapshot file

### LINKED ISSUE #281: Fetch snapshot files from external source

thibault-martinez:
This PR adds snapshot signature verification to the snapshot component.

Fixes #273 
Fixes #281 

# Test Plan:
Added and existing tests


common/model/bundle.h
```diff
@@ -13,6 +13,14 @@
 #include "common/trinary/flex_trit.h"
 #include "utarray.h"
 
+#define TRYTE_HASH_LENGTH 81
+#define TRIT_HASH_LENGTH 243
```
th0br0: let's call this `HASH_LENGTH_TRIT` and `HASH_LENGTH_TRYTE`? also, `HASH_LENGTH` is already defined elsewhere. Are you sure the other values aren't as well?

common/model/bundle.h
```diff
@@ -13,6 +13,14 @@
...
+#define MAX_TRYTE_VALUE 13
+#define MIN_TRYTE_VALUE -13
+#define TRYTE_WIDTH 3
+#define NUMBER_OF_SECURITY_LEVELS 3
```
th0br0: `SECURITY_LEVEL_MAX`?

common/model/bundle.h
```diff
@@ -13,6 +13,14 @@
 #include "common/trinary/flex_trit.h"
 #include "utarray.h"
 
+#define TRYTE_HASH_LENGTH 81
+#define TRIT_HASH_LENGTH 243
+#define MAX_TRYTE_VALUE 13
```
th0br0: `TRYTE_VALUE_MAX` & `_MIN`?

consensus/snapshot/snapshot.c
```diff
@@ -91,14 +92,19 @@ retcode_t iota_snapshot_init(snapshot_t *const snapshot,
...
 
-  // TODO signature validation
-  // String snapshotFile = config.getSnapshotFile();
-  // if (!config.isTestnet() &&
-  //     !SignedFiles.isFileSignatureValid(
-  //         snapshotFile, config.getSnapshotSignatureFile(), SNAPSHOT_PUBKEY,
-  //         SNAPSHOT_PUBKEY_DEPTH, SNAPSHOT_INDEX)) {
-  //   throw new RuntimeException("Snapshot signature failed.");
-  // }
+  if (!testnet) {
```
th0br0, line 221: testnet is also signed, we might want to have a separate flag for skipping the sig check?

utils/signed_files.c
```diff
@@ -0,0 +1,158 @@
...
+#include "common/curl-p/trit.h"
+#include "common/kerl/kerl.h"
+#include "common/model/bundle.h"
+#include "common/sign/v1/iss_curl.h"
+#include "common/trinary/trit_long.h"
+#include "common/trinary/tryte_ascii.h"
+#include "utils/merkle.h"
+#include "utils/signed_files.h"
+
+retcode_t is_file_signature_valid(char const *const filename,
```
th0br0: `iota_file_signature_validate` or similar?

utils/signed_files.c
```diff
@@ -0,0 +1,158 @@
...
+  flex_trit_t digest[FLEX_TRIT_SIZE_243];
+
+  if ((ret = digest_file(filename, digest))) {
+    return ret;
+  }
+  return validate_signature(signature_filename, public_key, depth, index,
+                            digest, valid);
+}
+
+retcode_t validate_signature(char const *const signature_filename,
```
th0br0: let's make this one static so we don't bloat the global function namespace?

utils/signed_files.c
```diff
@@ -0,0 +1,158 @@
...
+  *valid = (memcmp(public_key_trits, root, HASH_LENGTH) == 0);
+
+done:
+  if (fp) {
+    fclose(fp);
+  }
+  return ret;
+}
+
+retcode_t digest_file(char const *const filename, flex_trit_t *const digest) {
```
th0br0: `static`?

utils/signed_files.h
```diff
@@ -0,0 +1,41 @@
...
+#include "common/errors.h"
+#include "common/trinary/flex_trit.h"
+#include "common/trinary/trits.h"
+#include "common/trinary/tryte.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern retcode_t is_file_signature_valid(char const *const filename,
```
th0br0: no need for `extern`?

utils/tests/test_signed_files.c
```diff
@@ -0,0 +1,38 @@
...
+
+#include <unity/unity.h>
+
+#include "consensus/snapshot/snapshot.h"
+#include "utils/signed_files.h"
+
+void test_good_signature(void) {
+  bool valid = false;
+  TEST_ASSERT(
+      is_file_signature_valid(MAINNET_SNAPSHOT_FILE, MAINNET_SNAPSHOT_SIG_FILE,
```
th0br0: also test testnet?

th0br0 (changes requested):

consensus/snapshot/snapshot.c
thibault-martinez: I thought it was signed too, but IRI team told me it's not.
They pushed a testnet sigfile for consistency but it's not the actual signature.

utils/tests/test_signed_files.c
thibault-martinez: Ditto, not signed

common/model/bundle.h
thibault-martinez: These values are actually spread and repeated across many files. What about a common definitions file to centralize them ?

thibault-martinez: I'm doing it in another PR, it's becoming huge.

ciri/BUILD
```diff
@@ -1,19 +1,10 @@
...
-
 cc_binary(
     name = "ciri",
     srcs = ["main.c"],
     copts = ["-DLOGGER_ENABLE"],
-    data = [
-        "ciri.db",
-        ":snapshot_testnet",
-    ],
+    data = ["ciri.db"],
```
th0br0, line 7: Isn't this generated via some genrule elsewher? if so, we should reference the rule & not the file here

ciri/core.c
```diff
@@ -28,8 +27,16 @@ retcode_t core_init(core_t* const core) {
...
-                         CIRI_SNAPSHOT_SIG_FILE, core->config.testnet)) {
+
+  if (core->config.testnet) {
+    snapshot_file = TESTNET_SNAPSHOT_FILE;
+    snapshot_sig_file = TESTNET_SNAPSHOT_SIG_FILE;
+  } else {
+    snapshot_file = MAINNET_SNAPSHOT_FILE;
+    snapshot_sig_file = MAINNET_SNAPSHOT_SIG_FILE;
+  }
+  if (iota_snapshot_init(&core->snapshot, snapshot_file, snapshot_sig_file,
```
th0br0: add a newline above?

consensus/snapshot/BUILD
```diff
@@ -1,14 +1,35 @@
...
+    name = "snapshot_files",
+    outs = [
+        "snapshotMainnet.txt",
+        "snapshotMainnet.sig",
+        "snapshotTestnet.txt",
+        "snapshotTestnet.sig",
+    ],
+    cmd =
+        """
+        url="https://raw.githubusercontent.com/iotaledger/iri/master/src/main/resources"
```
th0br0: Could you replace the wgets etc with https://docs.bazel.build/versions/master/be/workspace.html#http_file ?

consensus/snapshot/snapshot.c
th0br0: :+1:

utils/signed_files.c
```diff
@@ -0,0 +1,173 @@
...
+    if ((trits = realloc(trits, HASH_LENGTH_TRIT *
+                                    (((read * 6) / HASH_LENGTH_TRIT) + 1))) ==
+        NULL) {
+      ret = RC_UTILS_OOM;
+      goto done;
+    }
+    memset(trits, 0, HASH_LENGTH_TRIT * (((read * 6) / HASH_LENGTH_TRIT) + 1));
+    trytes_to_trits(trytes, trits, read * 2);
+    kerl_absorb(&kerl, trits,
+                HASH_LENGTH_TRIT * (((read * 6) / HASH_LENGTH_TRIT) + 1));
```
th0br0, line 134: This is a memleak, you need to free `line`. Also: `This buffer should be freed by the user program even if getline() failed.`

utils/signed_files.c
```diff
@@ -0,0 +1,173 @@
...
+        sig_trits, RADIX * read, &curl);
+    curl_reset(&curl);
+  }
+  if (i != 3 || read < 0) {
+    ret = RC_UTILS_INVALID_SIG_FILE;
+    goto done;
+  }
+
+  iss_curl_address(sig_digests, root, sizeof(sig_digests), &curl);
+  if ((read = getline(&line, &len, fp)) > 0) {
```
th0br0, line 75: Memleak, you should free this. `This buffer should be freed by the user program even if getline() failed.`

th0br0 (changes requested):

utils/signed_files.c
thibault-martinez: `getline` realloc the `line` if necessary so we just have to free it once, and it's free'd L144. It's also free'd if `getline` failed but I just noticed that I'm not returning in this case, I'll fix that.

utils/signed_files.c
thibault-martinez: Ditto

consensus/snapshot/BUILD
thibault-martinez: This was very temporary, do we want to get this PR merged before the snapshot files repository exists ?

ciri/BUILD
thibault-martinez: It's not generated by a genrule. Hopefully this will be removed by #234

utils/signed_files.c
th0br0: The comment I `copied` above contradicts your `It's also free'd if getline failed?

thibault-martinez: Sorry my sentence wasn't clear, I meant in our case, we are freeing it, even if getline failed

utils/signed_files.c
thibault-martinez: Sorry, it was free'd in the other function, but not this one, fixing it rn

WORKSPACE
```diff
@@ -1,5 +1,36 @@
 workspace(name = "org_iota_entangled")
 
+SNAPSHOT_REPOSITORY = "https://raw.githubusercontent.com/thibault-martinez/snapshots/master"
```
th0br0: Could you move these into `tools/snapshot.bzl` and just load/execute from WORKSPACE so we don't clutter this file too much?

consensus/consensus.c
```diff
@@ -107,7 +111,8 @@ retcode_t iota_consensus_init(iota_consensus_t *const consensus,
 
   log_info(CONSENSUS_LOGGER_ID, "Initializing transaction validator\n");
   if ((ret = iota_consensus_transaction_validator_init(
-           &consensus->transaction_validator, SNAPSHOT_TIMESTAMP * 1000,
+           &consensus->transaction_validator,
+           consensus->snapshot.conf.timestamp * 1000,
```
th0br0: could we rename the variable to `timestampSec` so it's clear which resolution it has?
also, `1000uL`

consensus/snapshot/BUILD
```diff
@@ -1,15 +1,37 @@
+filegroup(
+    name = "snapshot_files",
```
th0br0, line 2: We should have a `config_setting` together with `select(`'s here to pick the right snapshot file group :thinking:

consensus/snapshot/BUILD
```diff
@@ -1,15 +1,37 @@
...
+    visibility = ["//visibility:public"],
+)
+
 cc_library(
     name = "snapshot",
     srcs = ["snapshot.c"],
     hdrs = ["snapshot.h"],
+    data = [":snapshot_files"],
+    defines = [
+        "SNAPSHOT_CONF_MAINNET='\"external/snapshot_conf_mainnet/file/snapshot.json\"'",
```
th0br0, line 19: same config_setting / select thing here, see how trit_encoding is setup for reference

consensus/snapshot/snapshot.c
```diff
@@ -76,9 +81,123 @@ static retcode_t iota_snapshot_initial_state(snapshot_t *const snapshot,
...
+  cJSON *json = NULL, *tmp = NULL, *timestamp = NULL, *signature = NULL,
+        *coordinator = NULL;
+
+  if ((file = fopen(snapshot_conf_file, "r")) == NULL) {
+    log_error(SNAPSHOT_LOGGER_ID, "Snapshot configuration file not found\n");
+    ret = RC_SNAPSHOT_FILE_NOT_FOUND;
+    goto done;
+  }
+
+  if (fseek(file, 0, SEEK_END) < 0 || (len = ftell(file)) < 0 ||
```
th0br0: shouldn't `ftell` be compared to `<=` 0?

consensus/snapshot/snapshot.c
```diff
@@ -76,9 +81,123 @@ static retcode_t iota_snapshot_initial_state(snapshot_t *const snapshot,
...
+    if (tmp == NULL || !cJSON_IsString(tmp) || tmp->valuestring == NULL ||
+        strlen(tmp->valuestring) != HASH_LENGTH_TRYTE) {
+      goto json_error;
+    }
+    memcpy(conf->signature_pubkey, tmp->valuestring, HASH_LENGTH_TRYTE);
+  }
+
+  goto done;
+
+json_error : {
```
th0br0, line 177: any reason for the braces?

consensus/snapshot/snapshot.h
```diff
@@ -37,26 +30,50 @@ typedef struct state_entry_t {
 
 typedef state_entry_t *state_map_t;
 
+typedef struct snapshot_conf_s {
+  uint64_t timestamp;
+  uint64_t signature_index;
+  uint64_t signature_depth;
+  tryte_t signature_pubkey[HASH_LENGTH_TRYTE];
```
th0br0: shouldn't we use flex_trits here?

th0br0 (changes requested):

consensus/snapshot/BUILD
thibault-martinez: But unlike trit_encoding, mainnet/testnet is a runtime option, not a compile-time option so we need to have both available ?

consensus/snapshot/snapshot.c
thibault-martinez: Indeed

consensus/snapshot/snapshot.h
thibault-martinez: It initially was, let me remember why I changed 😅

consensus/snapshot/snapshot.c
thibault-martinez: Labels can only point to statements or blocks.
So you have to put braces or do something like `json_error : ;` if you want to declare variables after a label.

consensus/snapshot/BUILD
thibault-martinez: As discussed, this will be handled in another PR where the network type would be a compile-time option.

consensus/snapshot/BUILD
thibault-martinez: As discussed, this will be handled in another PR where the network type would be a compile-time option.

consensus/snapshot/snapshot.h
thibault-martinez: done

consensus/exit_probability_randomizer/tests/test_exit_probability_randomizer.c
```diff
@@ -34,9 +34,9 @@ static char *test_db_path =
     "consensus/exit_probability_randomizer/tests/test.db";
 static char *ciri_db_path =
     "consensus/exit_probability_randomizer/tests/ciri.db";
-
 static char *snapshot_path =
     "consensus/exit_probability_randomizer/tests/snapshot.txt";
+static char *snapshot_conf_path = "consensus/snapshot/tests/snapshot_conf.txt";
```
th0br0: this file should end in .json?

th0br0 (approved):

consensus/exit_probability_randomizer/tests/test_exit_probability_randomizer.c
thibault-martinez: Yes

PR #317, opened 2018-10-01
Cclient API initial functions.

### LINKED ISSUE #55: Implement/Design `init_iota_http_service`

oopsmonk:
This PR fixed #55 
* Implemented `iota_client_core_init` and `iota_client_core_destroy` 
* Fixed `BUILD` dependence 
* Simplified marcos in json test cases. 

Example for using cclient core API:
```
int main(){
  retcode_t ret = RC_OK;
  iota_client_service_t serv;
  serv.http.host = "localhost";
  serv.http.port = 14265;
  serv.http.api_version = 1;
#ifdef _DEBUG_STDOUT_
  logger_init();
  logger_output_register(stdout);
  logger_output_level_set(stdout, LOGGER_DEBUG);
#endif
  iota_client_core_init(&serv);

  // emit getNodeInfo to IRI
  get_node_info_res_t* node_res = get_node_info_res_new();
  ret = iota_api_get_node_info(&serv, &node_res);
  get_node_info_res_free(&node_res);

  iota_client_core_destroy();
  return ret; 
}
```

# Test Plan:
CI must pass


cclient/BUILD
```diff
@@ -1,28 +1,28 @@
...
-    ]),
+    srcs = [
+        "iota_api.c",
+        "iota_api.h",
+    ],
     hdrs = ["iota_api.h"],
     visibility = ["//visibility:public"],
     deps = [
-        ":service",
+        "//cclient:service",
```
tsvisabo: why full path?

cclient/service.c
```diff
@@ -0,0 +1,24 @@
...
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "service.h"
+#include "serializer/json/json_serializer.h"
+#include "utils/logger_helper.h"
+
+void iota_client_service_init(iota_client_service_t* serv) {
+  // enable logger
+  logger_init_json_serializer();
```
tsvisabo: should be configurable?

tsvisabo (changes requested):

cclient/BUILD
```diff
@@ -1,28 +1,28 @@
 cc_library(
     name = "api",
-    srcs = glob([
-        "*.c",
-        "**/*.c",
-        "**/*.h",
-    ]),
+    srcs = [
+        "iota_api.c",
+        "iota_api.h",
```
thibault-martinez: Header in the sources ?

cclient/BUILD
```diff
@@ -1,28 +1,28 @@
...
+        "//cclient:service",
         "//cclient/http:beast",
     ],
 )
 
 cc_library(
     name = "service",
     srcs = [
+        "service.c",
         "service.h",
```
thibault-martinez: Header in the sources ?

cclient/service.h
```diff
@@ -20,27 +20,16 @@ extern "C" {
...
-typedef struct {
-  char* data;
-  size_t length;
-} iota_response_t;
+  http_info_t http;
+  serializer_t serializer;
+} iota_client_service_t;
 
+void iota_client_service_init(iota_client_service_t* serv);
+void iota_client_service_destroy();
```
thibault-martinez: newline

thibault-martinez (changes requested):

cclient/BUILD
oopsmonk: I thought it's no difference in this case, isn't it?

cclient/service.c
oopsmonk: not for now.

cclient/BUILD
oopsmonk: https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary.srcs

cclient/BUILD
oopsmonk: https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary.srcs

cclient/BUILD
thibault-martinez: Yes but you have them in srcs AND in hdrs so it's kind of redundant

cclient/BUILD
tsvisabo: exactly :) ,  but i think it's better to use `:` so people who read this file knows the rule is present

cclient/BUILD
oopsmonk: Ah, I got what you mean, thanks.

tsvisabo (approved):

thibault-martinez (approved):

th0br0:
We're probably missing a couple `const`ness modifiers here, but let's merge and revisit that later! Yay, CClient is usable :party_parrot:

PR #318, opened 2018-10-01
Fix entry point selector

### LINKED ISSUE #180: Implement entry point selector

thibault-martinez:
Fix some mistakes in the entry point selector.
Closes #180 

# Test Plan:
Added test


consensus/entry_point_selector/entry_point_selector.c
```diff
@@ -26,27 +26,31 @@ retcode_t iota_consensus_entry_point_selector_init(
...
-retcode_t iota_consensus_get_entry_point(entry_point_selector_t *const eps,
-                                         size_t const depth,
-                                         trit_array_p const ep) {
+retcode_t iota_consensus_entry_point_selector_get_entry_point(
+    entry_point_selector_t *const eps, size_t const depth,
+    trit_array_p const ep) {
   retcode_t ret = RC_OK;
-  size_t milestone_index = MAX(eps->mt->latest_milestone_index - depth - 1, 0);
+  uint64_t milestone_index =
+      MAX(eps->mt->latest_solid_subtangle_milestone_index - depth - 1, 0);
```
th0br0: ~I believe you're missing a cast to `int64_t` here?~ You're not, but we probably shouldn't use `size_t` for `latest_solid_subtangle_milestone_index`?

th0br0 (changes requested):

consensus/entry_point_selector/entry_point_selector.c
thibault-martinez: `latest_solid_subtangle_milestone_index` is an `uint64_t`

th0br0: Ah it's not on the `get_node_info` struct - unrelated.
So yeah, you definitely want to cast here?

tsvisabo: could use `ULLONG_MAX`

tsvisabo (changes requested):

tsvisabo (approved):

consensus/entry_point_selector/entry_point_selector.h
```diff
@@ -27,14 +27,42 @@ typedef struct entry_point_selector_s {
...
+ * Given a depth N, returns a N-deep milestone, i.e. (latest_solid - depth)th
+ * milestone.
+ *
+ * @param eps The entry point selector
+ * @param depth The depth
+ * @param ep The entry point
+ *
+ * @return a status code
+ */
+extern retcode_t iota_consensus_entry_point_selector_get_entry_point(
```
th0br0: Why the explicit extern? Because there can be multiple impls?

thibault-martinez: No specific reason, I saw it many times elsewhere and somehow thought it was part of the coding style, I'll remove.

th0br0: The motivation for that is link/build-time pluggable implementations. Might make sense here in the future as well, we'll see.

th0br0 (approved):

PR #329, opened 2018-10-02
Common definitions

thibault-martinez:
This PR 
- removes duplicate definitions of the same values
- renames values for clarity
- relocates all common definitions under `/common/defs.h`

# Test Plan:
CI

### DIFF
--- a/common/curl-p/hamming.c
+++ b/common/curl-p/hamming.c
@@ -5,11 +5,12 @@
  * Refer to the LICENSE file for licensing information
  */
 
-#include "hamming.h"
 #include <stdio.h>
-#include "ptrit.h"
-#include "search.h"
-#include "trit.h"
+
+#include "common/curl-p/hamming.h"
+#include "common/curl-p/ptrit.h"
+#include "common/curl-p/search.h"
+#include "common/curl-p/trit.h"
 
 short test(PCurl *const curl, unsigned short const security) {
   unsigned short i, j, k;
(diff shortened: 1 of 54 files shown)


th0br0:
Let's hold back on merging this until #297 is done as that also takes care of some #defines

common/defs.h
```diff
@@ -0,0 +1,32 @@
...
+#define NUMBER_OF_TRITS_IN_A_BYTE 5
+
+#define TRYTE_VALUE_MAX 13
+#define TRYTE_VALUE_MIN -13
+#define TRYTE_WIDTH 3
+
+#define HASH_LENGTH_BYTE 48
+#define HASH_LENGTH_TRYTE 81
+#define HASH_LENGTH_TRIT 243
+
```
tsvisabo, line 30: can you leave comments to distinguish between sections?

tsvisabo (changes requested):

common/defs.h
thibault-martinez: Sure

common/curl-p/hamming.c
```diff
@@ -18,7 +19,8 @@ short test(PCurl *const curl, unsigned short const security) {
     sum = 0;
 
     for (j = 0; j < security; j++) {
-      for (k = j * HASH_LENGTH / 3; k < (j + 1) * HASH_LENGTH / 3; k++) {
+      for (k = j * HASH_LENGTH_TRIT / 3; k < (j + 1) * HASH_LENGTH_TRIT / 3;
```
th0br0: While we're at it, we should also replace the various `HASH_LENGTH_TRIT / 3` with `HASH_LENGTH_TRYTE`

th0br0 (changes requested):

common/curl-p/hamming.c
thibault-martinez: Will do

th0br0 (approved):
Let's get this merged so it doesn't lag behind too much & we can reduce our backlog :)

tsvisabo (approved):

Issue #166, opened 2018-08-19
Efficient flex_trit_array_to_bytes


thibault-martinez:
In many cases we need to convert flex_trit arrays to bytes (e.g. converting a transaction to bytes to send it over network).

Current implementation is that we intermediary convert to trits.
`flex_trits -> trits -> bytes`

PR #346, opened 2018-10-03
common/trinary: bytes/flex_trits conversion, optimisations

ifullgaz:
This PR provides conversion functions between bytes and flex_trits - Issue #166.
Also, some optimisations to other conversion functions based work done in PR#310

# Test Plan:
Test plan as usual

### DIFF
--- a/common/trinary/flex_trit.c
+++ b/common/trinary/flex_trit.c
@@ -10,77 +10,81 @@
 
 #include "common/trinary/flex_trit.h"
 #include "common/trinary/trit_byte.h"
-
+#include "utils/macros.h"
 size_t flex_trits_slice(flex_trit_t *const to_flex_trits, size_t const to_len,
                         flex_trit_t const *const flex_trits, size_t const len,
                         size_t const start, size_t const num_trits) {
   // Bounds checking
-  if (num_trits > to_len || (start + num_trits) > len) {
+  if (num_trits == 0 || num_trits > to_len || (start + num_trits) > len) {
     return 0;
   }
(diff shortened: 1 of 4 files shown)


common/trinary/flex_trit.c
```diff
@@ -187,3 +198,107 @@ size_t flex_trits_from_trytes(flex_trit_t *to_flex_trits, size_t to_len,
...
+    offset -= trits_for_byte;
+  }
+#elif defined(FLEX_TRIT_ENCODING_5_TRITS_PER_BYTE)
+  size_t num_bytes = min_bytes(num_trits);
+  memcpy(to_flex_trits, bytes, num_bytes);
+  size_t residual = num_trits % 5;
+  if (residual) {
+    trit_t last_byte[5] = {0};
+    size_t index = num_bytes - 1;
+    byte_to_trits(bytes[index], last_byte, 4);
```
th0br0, line 299: Can't we just do `bytes[index] = FLEX_TRIT_NULL_VALUE` here?

th0br0: (Same goes for the other places?)

th0br0 (changes requested):

common/trinary/flex_trit.c
ifullgaz: Not really, the point is to leave only the last 'residual' trits from the last byte and eliminate the trailing trits. Also, bytes[index] contains n trits, not individual trits, so it makes little sense in this case to assign FLEX_TRIT_NULL_VALUE here.

th0br0 (approved):

PR #350, opened 2018-10-04
concensus/bundle_validator: fix CI tests failing

ifullgaz:
concensus/bundle_validator: fix CI tests failing

# Test Plan:
usual test suite

### DIFF
--- a/consensus/bundle_validator/bundle_validator.c
+++ b/consensus/bundle_validator/bundle_validator.c
@@ -175,6 +175,7 @@ retcode_t validate_signature(bundle_transactions_t* bundle,
   init_kerl(&address_kerl);
   init_kerl(&sig_frag_kerl);
 
+  *is_valid = true;
   for (curr_tx = (iota_transaction_t)utarray_eltptr(bundle, 0);
        curr_tx != NULL;) {
     if (curr_tx->value >= 0) {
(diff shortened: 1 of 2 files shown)


tsvisabo (approved):

consensus/bundle_validator/bundle_validator.c
```diff
@@ -38,15 +38,18 @@ retcode_t bundle_validate(const tangle_t* const tangle, trit_array_p tail_hash,
   retcode_t res = RC_OK;
   iota_transaction_t curr_tx;
 
+  *is_valid = true;
```
thibault-martinez, line 41: Isn't it easier/safer to initialise it to `false` ? Then you don't have to put `*is_valid = false;` on every error ?

th0br0: Whoops, sorry for resolving - mobile.
We.never set the value to true?

tsvisabo: yes @thibault-martinez , that's might actually be better, but i think for this PR we should leave it like that @ifullgaz  because the change is not that straight forward

thibault-martinez: In my head `res = validate_signature(bundle, normalized_bundle_trits, is_valid);` would have set the value to true, but it doesn't actually

ifullgaz: Yes, I just saw that as well. Will fix too.
As for setting is_valid to false, it's more difficult to set it to true later, because of the loop. Let's say that I like it's more readable IMMO.

PR #353, opened 2018-10-04
consensus/exit_probability_randomizer_test - 

tsvisabo:
Make test more conclusive (test that all tips were selected with the expected probability)
Reduce load


th0br0 (approved):

utils/macros.h
```diff
@@ -8,7 +8,18 @@
 #ifndef __UTILS_MACROS_H__
 #define __UTILS_MACROS_H__
 
-#define MAX(x, y) (((x) > (y)) ? (x) : (y))
-#define MIN(x, y) (((x) < (y)) ? (x) : (y))
+#define MAX(a, b)           \
+  ({                        \
+    __typeof__(a) _a = (a); \
```
thibault-martinez: Isn't this compiler-specific ?

tsvisabo: Yes, they are, i guess `type_of` is the correct one

th0br0:
#805 passed, looks like buildkite got a bit confused there.

Issue #361, opened 2018-10-07
byte_to_trits gives wrong results


ifullgaz:
byte_to_trits gives wrong results when i < index of highest radix index

PR #362, opened 2018-10-07
common/trinary: fix byte_to_trits gives wrong results

ifullgaz:
fix byte_to_trits gives wrong results (issue #361)
Adjust conversion functions to accept actual number of trits to convert instead of num - 1

# Test Plan:
usual test plan

### DIFF
--- a/common/trinary/trit_byte.c
+++ b/common/trinary/trit_byte.c
@@ -6,36 +6,35 @@
  */
 
 #include "trit_byte.h"
+#include "utils/macros.h"
+#include <assert.h>
 
 #define NUMBER_OF_TRITS_IN_A_BYTE 5
 #define RADIX 3
 static const byte_t byte_radix[] = {1, 3, 9, 27, 81};
 
-void trits_to_bytes(trit_t *trits, byte_t *bytes, size_t size) {
-  if (size <= 0) {
+void trits_to_bytes(trit_t *trits, byte_t *bytes, size_t num_trits) {
+  if (num_trits == 0) {
     return;
   }
-  size_t end;
-  if (size < NUMBER_OF_TRITS_IN_A_BYTE) {
-    end = size;
-  } else {
-    end = NUMBER_OF_TRITS_IN_A_BYTE;
-  }
-  bytes[0] = trits_to_byte(trits, 0, end - 1);
-  trits_to_bytes(&trits[end], &bytes[1], size - end);
+  size_t end = MIN(num_trits, NUMBER_OF_TRITS_IN_A_BYTE);
+  bytes[0] = trits_to_byte(trits, 0, end);
+  trits_to_bytes(&trits[end], &bytes[1], num_trits - end);
 }
 
-size_t min_bytes(size_t const trits_length) {
-  return (trits_length + NUMBER_OF_TRITS_IN_A_BYTE - 1) /
+size_t min_bytes(size_t const num_trits) {
+  return (num_trits + NUMBER_OF_TRITS_IN_A_BYTE - 1) /
          NUMBER_OF_TRITS_IN_A_BYTE;
 }
 
 byte_t trits_to_byte(trit_t const *const trits, byte_t const cum,
-                     size_t const i) {
-  if (i == 0) {
-    return cum * RADIX + trits[i];
+                     size_t const num_trits) {
+  assert(num_trits <= NUMBER_OF_TRITS_IN_A_BYTE);
+  if (num_trits == 0) {
+    return cum;
   }
-  return trits_to_byte(trits, cum * RADIX + trits[i], i - 1);
+  byte_t byte = cum * RADIX + trits[num_trits - 1];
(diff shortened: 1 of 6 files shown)


common/trinary/trit_byte.c
```diff
@@ -38,29 +38,37 @@ byte_t trits_to_byte(trit_t const *const trits, byte_t const cum,
...
+
 void bytes_to_trits(byte_t const *const byte, size_t const n_bytes,
                     trit_t *const trit, size_t const n_trits) {
   if (n_bytes == 0 || n_trits == 0) {
     return;
   }
   size_t end =
       n_trits < NUMBER_OF_TRITS_IN_A_BYTE ? n_trits : NUMBER_OF_TRITS_IN_A_BYTE;
-  byte_to_trits(*byte, trit, end - 1);
+  _byte_to_trits(*byte, trit, NUMBER_OF_TRITS_IN_A_BYTE - 1, end - 1);
```
tsvisabo: Can we have that being called non recursively?

common/trinary/trit_byte.c
```diff
@@ -38,29 +38,37 @@ byte_t trits_to_byte(trit_t const *const trits, byte_t const cum,
...
+  } else {
+    trit = 0;
+  }
+  if (j == i) {
+    trits[i] = trit;
+  }
+  if (j == 0) {
+    return;
+  }
+  _byte_to_trits(byte, trits, j - 1, j == i ? i - 1 : i);
```
tsvisabo: The recursive stack here can't be that bad because it can contain 5 calls max, but would be better to have its iterative version

common/trinary/trit_byte.c
```diff
@@ -38,29 +38,37 @@ byte_t trits_to_byte(trit_t const *const trits, byte_t const cum,
...
+void _byte_to_trits(byte_t byte, trit_t *const trits, size_t const j, size_t const i) {
+  trit_t trit;
+  if (byte > (byte_radix[j] >> 1)) {
+    byte -= byte_radix[j];
+    trit = 1;
+  } else if (byte < -(byte_radix[j] >> 1)) {
+    byte += byte_radix[j];
+    trit = -1;
+  } else {
+    trit = 0;
```
tsvisabo: can you initialize trit and get rid of the else?

common/trinary/trit_byte.c
```diff
@@ -38,29 +38,37 @@ byte_t trits_to_byte(trit_t const *const trits, byte_t const cum,
   return trits_to_byte(trits, cum * RADIX + trits[i], i - 1);
 }
 
+void _byte_to_trits(byte_t byte, trit_t *const trits, size_t const j, size_t const i) {
```
tsvisabo: Do we really need last parameter `i`?
If `bytes_to_trits` is called with *n_trits == 1* doesn't it mean that remaining trits are all 0 anyways?
If so, you can always do: `trits[i] = trit;` and get same results, make sense?

tsvisabo (changes requested):

common/trinary/trit_byte.c
ifullgaz: Yes, it's recursive in _byte_to_trits. I haven't changed the original general algorithm

tsvisabo: ok, let's open an issue for that

common/trinary/trit_byte.c
ifullgaz: Done

common/trinary/trit_byte.c
ifullgaz: Done

common/trinary/trit_byte.c
ifullgaz: Yes, because we have to start converting with the highest order index (j) but only get the trits for the lower order. Potentially, i can be less than j on initial call (it happens in flex_trit)

common/trinary/trit_byte.c
```diff
@@ -37,25 +37,22 @@ byte_t trits_to_byte(trit_t const *const trits, byte_t const cum,
...
-  if (byte > (byte_radix[j] >> 1)) {
-    byte -= byte_radix[j];
-    trit = 1;
-  } else if (byte < -(byte_radix[j] >> 1)) {
-    byte += byte_radix[j];
-    trit = -1;
-  } else {
-    trit = 0;
+void _byte_to_trits(byte_t byte, trit_t *const trits, size_t j, size_t i) {
+  while (j < -1) {
```
tsvisabo, line 41: nice, but hacky, can you just >= 0 ?

tsvisabo (changes requested):

common/trinary/trit_byte.c
ifullgaz: No because j is a size_t and therefore is always >= 0. -1 automatically takes the highest value, which will never be the requested size

tsvisabo (approved):

PR #26, opened 2018-06-23
Add build rule for iOS static frameworks

rajivshah3:
See https://gitlab.com/iota-foundation/software/entangled/merge_requests/41

# Test Plan:
Tests can be run with:
```
xcodebuild test -project mobile/ios/Demo/Demo.xcodeproj -scheme EntangledDemo -destination 'platform=iOS Simulator,name=iPhone 6s'
```

PR #367, opened 2018-10-07
Change Android method names to match helpers

rajivshah3:
@th0br0 mentioned in an earlier conversation that the Android method names should be updated to match https://github.com/iotaledger/entangled/tree/develop/common/helpers. Additionally, this will allow the Android methods to be the same as the iOS methods (#26), which will make things more convenient for developers using both libraries (e.g. https://github.com/iotaledger/iota.js/pull/286)

# Test Plan:
`bazel build //mobile/android:jni` should succeed


th0br0:
Did I now? 🤔 
Thanks for the chore!

th0br0 (approved):

PR #379, opened 2018-10-09
consensus/exit_prob_randomizer/cw - use queue, returned value check

tsvisabo:


consensus/cw_rating_calculator/cw_rating_dfs_impl.c
```diff
@@ -160,11 +159,13 @@ static retcode_t cw_rating_dfs_do_dfs_from_db(
           return res;
         }
       }
+      continue;
     }
+    hash_stack_pop(&stack);
```
thibault-martinez, line 164: You have two `hash_stack_pop(&stack);` ?

thibault-martinez (changes requested):

consensus/cw_rating_calculator/cw_rating_dfs_impl.c
tsvisabo: but only one will be executed since there's `continue;`

thibault-martinez: Alright, been confused by the brackets

thibault-martinez (approved):

Issue #381, opened 2018-10-10
All consensus variables should be configurable with CLI arguments


thibault-martinez:
e.g. `MAX_ANALYZED_TXS`, `MAX_DEPTH`, `ALPHA`...

PR #382, opened 2018-10-10
Consensus init/start/stop/destroy

### LINKED ISSUE #337: Global init function for all consensus modules

thibault-martinez:
This PR adds a way to init, start, stop and destroy all consensus components as a whole.
Fixes #337 

# Test Plan:
Not applicable


tsvisabo (changes requested):
Can you also check if you can use `consensus_init` in the `exit_probability_*` tests instead of `init_epv`?

th0br0:
:thinking: instead of that `testnet` flag we should much rather hand a `network_config_t`  struct around that is setup with the various parameters we care about (req hash length, mwm, ...) but that's a separate issue. lgtm otherwise

thibault-martinez:
Agreed, I opened https://github.com/iotaledger/entangled/issues/381 earlier for that

thibault-martinez:
@tsvisabo I can, but I wouldn't be able to set `mt.latest_snapshot->index = 99999999999;`, or set it for every test.
Ok, nvm

tsvisabo (approved):

PR #395, opened 2018-10-14
utils/hash_structs - hash_set_for_each

tsvisabo:

### DIFF
--- a/utils/hash_containers.c
+++ b/utils/hash_containers.c
@@ -128,6 +128,19 @@ void hash_set_free(hash_set_t *set) {
   *set = NULL;
 }
 
+retcode_t hash_set_for_each(hash_set_t *set, hash_on_container_func func,
+                            void *container) {
+  retcode_t ret = RC_OK;
+  hash_set_entry_t *curr_entry = NULL;
+  hash_set_entry_t *tmp_entry = NULL;
+  HASH_ITER(hh, *set, curr_entry, tmp_entry) {
+    if ((ret = func(container, curr_entry->hash))) {
+      return ret;
+    }
+  }
+  return ret;
+}
+
 /*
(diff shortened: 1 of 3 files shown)


utils/hash_containers.c
```diff
@@ -128,6 +128,19 @@ void hash_set_free(hash_set_t *set) {
   *set = NULL;
 }
 
+retcode_t hash_set_for_each(hash_set_t *set, hash_on_container_func func,
+                            void *container) {
+  retcode_t ret = RC_OK;
+  hash_set_entry_t *curr_entry = NULL;
```
th0br0, line 134: As we call a `bitset` a `bitset`, we should probably call this `hashset` as well? (just a side remark)

utils/hash_containers.c
```diff
@@ -128,6 +128,19 @@ void hash_set_free(hash_set_t *set) {
...
+  hash_set_entry_t *tmp_entry = NULL;
+  HASH_ITER(hh, *set, curr_entry, tmp_entry) {
+    if ((ret = func(container, curr_entry->hash))) {
```
th0br0: This makes the implicit assumption that `RC_OK == 0`.

th0br0 (changes requested):

utils/hash_containers.c
tsvisabo: yeah, we can try to explicitly use `RC_OK`, should be enforced anywhere though

utils/hash_containers.c
tsvisabo: Sounds good for set, but then we have hashqueue,hashlist,hashdll (hashdoublelinkedlist?)

th0br0 (approved):

PR #399, opened 2018-10-15
cclient/http: replaced beast.cc with http.c

ifullgaz:
cclient/http: replaced beast.cc with http.c, a socket based implementation of the client

# Test Plan:
Usual test plan


cclient/BUILD
```diff
@@ -7,7 +7,7 @@ cc_library(
     visibility = ["//visibility:public"],
     deps = [
         ":service",
-        "//cclient/http:beast",
+        "//cclient/http:http",
```
th0br0, line 14: just `"//cclient/http"`

cclient/http/http.c
```diff
@@ -0,0 +1,231 @@
...
+  }
+  return RC_OK;
+}
+
+static retcode_t send_headers_to_iota_service(int sockfd,
+                                              http_info_t const* http_settings,
+                                              size_t data_length,
+                                              char const* const path) {
+  char header[100];
+  sprintf(header, "POST %s HTTP/1.1\r\n", path);
```
th0br0: We should send all of this message at once and not construct piecewise like this

cclient/http/http.c
```diff
@@ -0,0 +1,231 @@
...
+  if (send_data_to_iota_service(sockfd, header, strlen(header)) < 0) {
+    return -1;
+  }
+  if (send_data_to_iota_service(sockfd, "\r\n", 2) < 0) {
+    return -1;
+  }
+  return RC_OK;
+}
+
+retcode_t _iota_service_query(const void* const service_opaque,
```
th0br0: if it's private, this should be static. else why is there an underscore before the method name?

utils/sockets.h
```diff
@@ -0,0 +1,36 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Windows headers
+#ifdef __WIN32__
+#include <winsock2.h>
+#include <ws2tcpip.h>
+// ESP headers
+#elif __XTENSA__
```
th0br0: Isn't there a better, more LWIP-specific include guard?

cclient/http/tests/test_http.c
```diff
@@ -0,0 +1,139 @@
...
+    "Suspendisse laoreet. Fusce ut est sed dolor gravida convallis. Morbi "
+    "vitae ante. Vivamus ultrices luctus nunc. Suspendisse et dolor. Etiam "
+    "dignissim. Proin malesuada adipiscing lacus. Donec metus. Lorem ipsum "
+    "dolor sit amet, consectetuer adipiscing elit. Phasellus hendrerit. "
+    "Pellentesque aliquet nibh nec urna. In nisi neque, aliquet vel, dapibus "
+    "id, mattis vel, nisi. Sed pretium, ligula sollicitudin laoreet viverra, "
+    "tortor libero sodales leo, eget blandit nunc tortor eu nibh. Nullam "
+    "mollis. Ut justo. Suspendisse potenti. Sed egestas, ante et vulputate "
+    "volutpat, eros pede semper est, vitae luctus metus libero eu augue. Morbi "
+    "purus libero, faucibus adipiscing, commodo quis, gravida id, est. Sed "
```
th0br0, line 84: do we need the full lorem ipsum?

cclient/http/http.c
```diff
@@ -0,0 +1,231 @@
...
+ */
+
+#include "http.h"
+#include <string.h>
+#include "cclient/service.h"
+#include "http_parser.h"
+#include "utils/macros.h"
+#include "utils/sockets.h"
+
+#define RECEIVE_BUFFER_SIZE 1024
```
th0br0: that's pretty low. we should go for a larger buffer given that a tx is already 2680 bytes

cclient/http/http.c
th0br0: this should alos be static?

cclient/http/tests/test_http.c
```diff
@@ -0,0 +1,139 @@
...
+    "Pellentesque adipiscing eros ut libero. Ut condimentum mi vel tellus. "
+    "Suspendisse laoreet. Fusce ut est sed dolor gravida convallis. Morbi "
+    "vitae ante. Vivamus ultrices luctus nunc. Suspendisse et dolor. Etiam "
+    "dignissim. Proin malesuada adipiscing lacus.";
+
+retcode_t _iota_service_query(const void* const service_opaque,
+                              char_buffer_t* obj, char_buffer_t* response,
+                              char const* const path);
+
+void test_http(void) {
```
th0br0, line 110: this test relies on an external service. could you mark it as flaky in bazel so it won't crash the build if httpbin goes offline?

th0br0 (changes requested):

cclient/BUILD
ifullgaz: Ok

cclient/http/http.c
ifullgaz: Ok

cclient/http/http.c
ifullgaz: It's only used for testing, because of the `path` parameter which is not needed for our calls (it defaults to "/"). If I make this function static, it won't be available from the test code, which uses a different value for `path`. Not ideal, but close to a "module private function"

utils/sockets.h
ifullgaz: I haven't found one yet. I can't test for lwip defined macros at this point since lwip hasn't been included yet, so it's per platform really. At present, we have only discussed the Espressif SDK, which define `__XTENSA__`. Other platforms or SDKs will define their own.

cclient/http/tests/test_http.c
ifullgaz: I wanted to test the proper functioning of the `read_data_from_iota_service` function, which might receive data in chunks. The test server echos whatever is being sent to it, so the longer the text, the better the test. Unless the `read_data_from_iota_service` code is not going to change, I suggest keeping a long text here.

cclient/http/http.c
ifullgaz: Ok, but just guessing here. I've moved the `#define` to `utils/sockets.h`. A better solution would be to pickup the current socket receive size from the system at run time and allocate the buffer dynamically. The problem: MSVC doesn't support VLA so we need to use `malloc` - but maybe it would be acceptable to `#ifdef` the code just for Windows?

cclient/http/tests/test_http.c
ifullgaz: Ok

cclient/http/http.c
```diff
@@ -0,0 +1,218 @@
...
+                              char const* const path) {
+  int sockfd = -1;
+  retcode_t result = RC_OK;
+  const iota_client_service_t* const service =
+      (const iota_client_service_t* const)service_opaque;
+  const http_info_t* http_settings = &service->http;
+
+  if ((sockfd = connect_to_iota_service(http_settings->host,
+                                        http_settings->port)) < 0) {
+    return -1;
```
oopsmonk: should return retcode_t, and line number 199, 203, 207.

ifullgaz: Ok

WORKSPACE
```diff
@@ -2,7 +2,7 @@ workspace(name = "org_iota_entangled")
 
 git_repository(
     name = "rules_iota",
-    commit = "8571197145287c75a2b9cd93531db27d765fafc4",
+    commit = "dd8661f9728ca1dc8484bf90797e56f200c969b6",
```
th0br0: update to latest rules_iota ?

th0br0: `893bc942f22aa6ad5bbd72e7d86c94452fbd76d8`

th0br0 (changes requested):
Would it be possible to split the actual network/socket operation & http into separate interfaces?

cclient/http/http.c
th0br0: Agree. We've started using malloc in a lot of other places anyway, it's sane to do that here as well I guess. We'll pay the overhead of that later if need be.

th0br0: Could you file an issue for getting a better receive buffer size dynamically?

cclient/http/http.c
```diff
@@ -0,0 +1,183 @@
...
+
+static retcode_t send_headers_to_iota_service(int sockfd,
+                                              http_info_t const* http_settings,
+                                              size_t data_length) {
+  static char* header_template =
+      "POST %s HTTP/1.1\r\n"
+      "Host: %s\r\n"
+      "X-IOTA-API-Version: %d\r\n"
+      "Content-Type: application/x-www-form-urlencoded\r\n"
+      "Content-Length: %lu\r\n"
```
th0br0, line 147: The HTTP request header should specify the accepted return values (in our case `Accept: application/json\r\n`)

th0br0 (changes requested):
Just a nit

cclient/http/http.c
ifullgaz: Ok

th0br0 (approved):

PR #408, opened 2018-10-17
ledger_validator: clean and add state deltas

thibault-martinez:
Clean and finish the ledger validator.
Added state deltas and missing things.

# Test Plan:
CI must pass


consensus/ledger_validator/ledger_validator.c
```diff
@@ -85,44 +88,58 @@ static retcode_t build_snapshot(ledger_validator_t *const lv,
...
+                                  bool *const valid_delta) {
   retcode_t ret = RC_OK;
-  int number_of_analyzed_transactions = 0;
   hash_stack_t non_analyzed_hashes = NULL;
+  hash_set_t counted_hashes = NULL;
   iota_transaction_t tx_bundle = NULL;
   DECLARE_PACK_SINGLE_TX(tx, tx_ptr, pack);
-  struct _trit_array tx_hash = {NULL, NUM_TRITS_HASH, FLEX_TRIT_SIZE_243, 0};
+  struct _trit_array curr_hash = {NULL, NUM_TRITS_HASH, FLEX_TRIT_SIZE_243, 0};
+  flex_trit_t null_hash[FLEX_TRIT_SIZE_243];
```
tsvisabo: this should be defined as static and initialized outside the context of this function,
we have several places already where we have that (`genesis_hash`)
maybe add a struct with definitions that will be initialized via `consensus_init`?

tsvisabo (changes requested):

consensus/ledger_validator/ledger_validator.c
thibault-martinez: We have `common/defs.h`, I can add `common/defs.c` as well for variables like that.

utils/hash_containers.c
```diff
@@ -102,14 +104,14 @@ retcode_t hash_set_add(hash_set_t *const set, flex_trit_t const *const hash) {
...
+retcode_t hash_set_append(hash_set_t const *const set_from,
+                          hash_set_t *const set_to) {
   retcode_t ret = RC_OK;
   hash_set_entry_t *iter = NULL, *tmp = NULL;
 
-  HASH_ITER(hh, *set1, iter, tmp) {
-    if (!hash_set_contains(set2, iter->hash)) {
-      if ((ret = hash_set_add(set2, iter->hash)) != RC_OK) {
+  HASH_ITER(hh, *set_from, iter, tmp) {
+    if (!hash_set_contains(set_to, iter->hash)) {
```
th0br0: If it's a set you shouldn't have to check if it contains it already

th0br0 (changes requested):

utils/hash_containers.c
thibault-martinez: `hash_set_add` could take a bool and set the value if it added or not ?

th0br0: don't think we care about whether it was added right now? but ya, could do that.

thibault-martinez: I didn't see you were talking of `hash_set_append ` with the little piece of code.
You're right, but utash set does not check for unicity, we have to do it ourselves.
Anyway, this `hash_set_contains` should go to `hash_set_add` instead.

consensus/ledger_validator/ledger_validator.c
```diff
@@ -85,44 +88,58 @@ static retcode_t build_snapshot(ledger_validator_t *const lv,
...
+                                            &delta)) != RC_OK) {
+      goto done;
+    }
+    if (delta != NULL && !state_delta_empty(delta)) {
+      if ((ret = iota_snapshot_create_patch(
+               lv->milestone_tracker->latest_snapshot, &delta, &patch)) !=
+          RC_OK) {
+        goto done;
+      }
+      if (state_delta_is_consistent(&patch)) {
```
tsvisabo, line 98: don't we want to return error/log when patch is not consistent?

tsvisabo (changes requested):

consensus/ledger_validator/ledger_validator.c
thibault-martinez: We should definitely log something.
I'm not sure if we should return an error though, I'll investigate.

thibault-martinez: Yes, definitely an error, done.

consensus/defs.c
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "consensus/defs.h"
+
+flex_trit_t genesis_hash[FLEX_TRIT_SIZE_243] = {0};
```
th0br0: No globals please? If need be, let's make this a #define, but I'd much rather see this live in some state object

th0br0 (changes requested):

tsvisabo (approved):

th0br0 (approved):

PR #411, opened 2018-10-18
api: logic proposal

thibault-martinez:
This PR proposes and opens a discussion about the API logic design.

# Test Plan:
Not applicable.


ciri/api/BUILD
```diff
@@ -5,6 +5,8 @@ cc_library(
     srcs = ["api.c"],
     hdrs = ["api.h"],
     deps = [
+        "//cclient/request:requests",
```
tsvisabo, line 8: should we move `request/response` to a neutral namespace and use it as a dependency on both cIRI/cclient ?

tsvisabo (changes requested):
LGTM

ciri/api/BUILD
thibault-martinez: I think so, and serialization too.
Any suggestion where ?

tsvisabo: `common/network`  ?

tsvisabo (approved):

ciri/api/BUILD
th0br0: For the record: we decided to hold off on moving this for now

ciri/api/api.c
```diff
@@ -7,42 +7,212 @@
...
+  return RC_OK;
+}
+
+static retcode_t get_transactions_to_approve(
+    get_transactions_to_approve_req_t const *const req,
+    get_transactions_to_approve_res_t *const res) {
+  return RC_OK;
+}
+
+static retcode_t attach_to_tangle(attach_to_tangle_req_t const *const req,
```
th0br0: These methods should be able to be called directly.
- tests need this
- a custom IRI node with user code might want to call them without going through the API overhead

th0br0 (changes requested):

ciri/api/api.c
thibault-martinez: Alright, I was wondering how to test it.

thibault-martinez: Done.

th0br0 (approved):

PR #439, opened 2018-10-23
Hash containers refactoring WIP

thibault-martinez:
This PR introduces hash containers for all usual sizes, generated with Bazel.

# Test Plan:
Not applicable.


utils/containers/sets/hash_set_template.h
```diff
@@ -0,0 +1,46 @@
+/*
```
th0br0: Could we call this file `hash_set.h.tpl`?

utils/containers/sets/hash_set_template.c
```diff
@@ -0,0 +1,76 @@
+/*
```
th0br0: same filename convention here so it doesn't get globbed accidentally?

utils/containers/sets/hash_set_generator.bzl
```diff
@@ -0,0 +1,26 @@
+def hash_set_generate(size):
```
th0br0: Instead of relying on sed, could you use 
an approach similar to https://github.com/bazelbuild/examples/blob/master/rules/expand_template/hello.bzl so that we're OS-agnostic? docs are at https://docs.bazel.build/versions/master/skylark/lib/actions.html#expand_template

th0br0 (changes requested):

th0br0 (approved):

PR #440, opened 2018-10-24
transaction_requester: refactoring

### LINKED ISSUE #344: Refactor transaction requester to priorize milestones

thibault-martinez:
- Introduces milestone requesting (fixes #344)
- Use `hash243_set_t` + `rw_lock_handle_t`

# Test Plan:
CI must pass


gossip/components/requester.c
```diff
@@ -17,136 +16,184 @@
...
 
 bool requester_is_full(requester_state_t *const state) {
+  size_t size = 0;
+
   if (state == NULL) {
     return RC_REQUESTER_COMPONENT_NULL_STATE;
   }
 
-  return CL_SIZE(state->list) >= MAX_TX_REQ_NBR;
+  rw_lock_handle_rdlock(&state->lock);
```
tsvisabo: `Thread safety`
`You can use uthash in a threaded program. But you must do the locking. Use a read-write lock to protect against concurrent writes. It is ok to have concurrent readers (since uthash 1.5).`

tsvisabo: we are on 2.0.2

gossip/components/requester.c
```diff
@@ -17,136 +16,184 @@
...
-                                     trit_array_p *hash) {
+                                     flex_trit_t *const hash,
+                                     bool const milestone) {
   retcode_t ret = RC_OK;
-  concurrent_list_node_trit_array_p *iter = NULL;
+  hash243_set_t request_set = NULL;
+  hash243_set_t backup_set = NULL;
+  hash243_set_entry_t *iter = NULL;
+  hash243_set_entry_t *tmp = NULL;
+  trit_array_t key = {.trits = hash,
```
tsvisabo: Should assign `.trits = NULL`

tsvisabo (changes requested):

gossip/neighbor.c
```diff
@@ -77,16 +77,14 @@ retcode_t neighbor_send(node_t *const node, neighbor_t *const neighbor,
...
   }
-  if (hash != NULL) {
-    // TODO(thibault): iota_packet_set_request
-    flex_trits_to_trits(hash_trits, NUM_TRITS_HASH, hash->trits,
-                        hash->num_trits, hash->num_trits);
-    trits_to_bytes(hash_trits, packet->content + PACKET_TX_SIZE,
-                   NUM_TRITS_HASH);
-  }
+  // TODO(thibault): iota_packet_set_request
+  flex_trits_to_trits(hash_trits, HASH_LENGTH_TRIT, hash, HASH_LENGTH_TRIT,
```
tsvisabo: what if hash is the "null hash" ?

tsvisabo (changes requested):

gossip/components/requester.c
thibault-martinez: I'm actually wrapping the `flex_trit_t *` in a `trit_array_t` to do the database request

gossip/neighbor.c
thibault-martinez: If it's the null hash, it means that there are no transaction left to be requested but we still need to send that packet so I guess null hash is the way to go... ?

gossip/components/requester.c
thibault-martinez: After discussion/comprehension of a pretty elusive documentation, we'll let the `rw_lock_handle_rdlock `.

gossip/components/requester.c
tsvisabo: i see, but you don't need that first assignment because of line 179,
just so it will be little more clear

tsvisabo (changes requested):

gossip/components/requester.c
thibault-martinez: You are indeed totally right, I thought you were talking about `request_transaction`, I'll change that.

tsvisabo (approved):

th0br0 (approved):

tsvisabo (approved):

PR #445, opened 2018-10-26
consensus/transaction_solidifier - update solid states with one state…

### LINKED ISSUE #443: Update transaction solid state - multiple rows at once

tsvisabo:
Fixes #443


common/storage/sql/sqlite3/storage.c
```diff
@@ -404,6 +419,40 @@ retcode_t iota_stor_transaction_update_solid_state(
   return ret;
 }
 
+retcode_t iota_stor_transactions_update_solid_state_true(
+    const connection_t* const conn, const hash243_set_t hashes) {
```
th0br0: Can't we encode the value to set as an argument? Having a _true (& _false) method sounds nasty - the overhead is also negligible

th0br0 (changes requested):

common/storage/sql/sqlite3/storage.c
tsvisabo: sure

common/storage/sql/sqlite3/storage.c
```diff
@@ -404,6 +452,49 @@ retcode_t iota_stor_transaction_update_solid_state(
...
+  sqlite3_stmt* sqlite_statement = NULL;
+
+  if ((ret = begin_transaction((sqlite3*)conn->db)) != RC_OK) {
+    return ret;
+  }
+
+  if ((ret = prepare_statement(
+           (sqlite3*)conn->db, &sqlite_statement,
+           iota_statement_transaction_update_solid_state)) != RC_OK) {
+    goto done;
```
th0br0, line 470: in this case, we shouldn't rollback.

th0br0 (changes requested):

th0br0 (approved):

PR #470, opened 2018-11-04
Processor refactoring

thibault-martinez:
- Removes a useless dynamic allocation that caused a leak
- Adds transaction validation
- Changes the queue to a utash-based queue + cond/lock
- Removes dependency to core
- Some performance improvements
- Renames/cleans/documents

# Test Plan:
CI must pass.

PR #260, opened 2018-09-07
ciri: exit on CTRL-C

howjmay:
call core_stop(&core_g) and core_destroy(&core_g) if pressed ctrl-C, close #133

PR #472, opened 2018-11-05
main: print tx stats log

thibault-martinez:
The main is now printing tx stats logs.

Sample: `2018-11-05 12:04:21:           main:   INFO: Transactions: to process 0, to broadcast 0, to request 49, count 3956`

Depends on #470, will rebase after merge.

Should make #260 easier to implement.

# Test Plan:
Not applicable.

### DIFF
--- a/ciri/main.c
+++ b/ciri/main.c
@@ -13,6 +13,7 @@
 #include "utils/logger_helper.h"
 
 #define MAIN_LOGGER_ID "main"
+#define STATS_LOG_INTERVAL 10
 
 static core_t core_g;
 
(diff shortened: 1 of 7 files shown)


ciri/main.c
```diff
@@ -65,7 +66,22 @@ int main(int argc, char* argv[]) {
   flex_trit_t dummy[FLEX_TRIT_SIZE_8019];
   memset(dummy, FLEX_TRIT_NULL_VALUE, FLEX_TRIT_SIZE_8019);
   broadcaster_on_next(&core_g.node.broadcaster, dummy);
-  sleep(1000);
+
+  size_t count = 0;
+  while (true) {
```
th0br0, line 75: This breaks shutdown because it won't ever be interrupted?

th0br0 (changes requested):

ciri/main.c
thibault-martinez: It temporarily replaces a `sleep` with an arbitrary value so until #260 is merged we won't really have a sane shutdown.

thibault-martinez: We can wait until #260 is merged to merge this one.

th0br0 (approved):

PR #476, opened 2018-11-05
trinary: fix slice test segmentation fault

thibault-martinez:
# Test Plan:
Trinary tests must pass


th0br0 (approved):

thibault-martinez:
Seems like with this I avoided the cause, but not solved it. Still investigating.

th0br0 (changes requested):
Doesn't this also potentially lead to bad data?

thibault-martinez:
The root cause has been identified after a very productive call with @ifullgaz :)

th0br0 (approved):

PR #483, opened 2018-11-06
mam/v2: complete test suite

tsvisabo:


jserv:
Typo. Should be test suite.

PR #484, opened 2018-11-07
Responder refactoring

thibault-martinez:
- Changes queue to utash-based + cond/lock
- Removes useless dynamic allocations & leaks
- Uses `flex_trit_t *` instead of `trit_array_p`
- Adds `p_propagate_request`
- Adds `p_reply_random_tip`
- Adds `p_send_milestone`
- Nits & clean
- Documentation
- Adds a rand handle

Still working on random tip request

# Test Plan:
CI must pass


gossip/components/responder.c
```diff
@@ -69,11 +73,13 @@ static retcode_t reply_to_request(responder_t *const processor,
     }
   } else {
     // Transaction not found
-    // TODO(thibault): Randomly doesn't propagate request
-    if (!flex_trits_is_null(request->hash, FLEX_TRIT_SIZE_243)) {
+    if (!flex_trits_is_null(request->hash, FLEX_TRIT_SIZE_243) &&
```
tsvisabo: can you rename `flex_trits_is_null` to `flex_trits_are_null`?

tsvisabo (approved):

gossip/components/responder.c
```diff
@@ -6,219 +6,280 @@
...
-  if ((ret = iota_tangle_transaction_load(state->tangle, TRANSACTION_FIELD_HASH,
-                                          request->hash, &pack))) {
-    return ret;
+  // If the hash is null, a random tip was requested
+  if (flex_trits_is_null(hash->trits, FLEX_TRIT_SIZE_243)) {
+    log_debug(RESPONDER_LOGGER_ID, "Responding to random tip request\n");
+    if (true) {
+      neighbor->nbr_random_tx_req++;
+    }
+    // Else no tx to request, so no random tip will be sent as a reply.
```
th0br0, line 50: add a todo to take reply_random_tip into account here?

gossip/components/responder.c
```diff
@@ -6,219 +6,280 @@
...
   } else {
-    // Transaction not found
-    // TODO(thibault): Randomly doesn't propagate request
-    if (trit_array_is_null(request->hash) == false) {
-      // Request is an actual missing transaction
-      if ((ret = request_transaction(&state->node->transaction_requester,
-                                     request->hash->trits, false)) != RC_OK) {
+    // If a transaction was requested but not found, requests it
+    if (!flex_trits_is_null(hash, FLEX_TRIT_SIZE_243) &&
+        ((double)rand() / (double)RAND_MAX) <
```
th0br0: Could we abstract rand() please? (so other impls can use a more secure (p)RNG or simialr)

gossip/components/responder.c
```diff
@@ -6,219 +6,280 @@
...
+            responder->node->conf.p_propagate_request) {
```
th0br0: The `p_propagate_request` check should happen before we do or attempt any tx loading

gossip/components/responder.c
```diff
@@ -6,219 +6,280 @@
...
+    if (get_transaction_for_request(responder, request->neighbor, &hash,
+                                    &pack) != RC_OK) {
+      log_warning(RESPONDER_LOGGER_ID,
+                  "Getting transaction for request failed\n");
+      continue;
+    }
+    if (respond_to_request(responder, request->neighbor, request->hash,
+                           &pack) != RC_OK) {
+      log_warning(RESPONDER_LOGGER_ID, "Replying to request failed\n");
+      continue;
```
th0br0: you don't unlock here.

gossip/components/responder.c
```diff
@@ -6,219 +6,280 @@
...
+      rw_lock_handle_unlock(&responder->lock);
+      continue;
+    }
+    log_debug(RESPONDER_LOGGER_ID, "Responding to request\n");
+    hash_pack_reset(&pack);
+    hash.trits = request->hash;
+    if (get_transaction_for_request(responder, request->neighbor, &hash,
+                                    &pack) != RC_OK) {
+      log_warning(RESPONDER_LOGGER_ID,
+                  "Getting transaction for request failed\n");
```
th0br0, line 160: you don't unlock here

th0br0 (changes requested):

gossip/components/responder.c
thibault-martinez: But it's the probability of propagating a request if a requested transaction wasn't found, so we have to try to load the transaction first ?

gossip/components/responder.c
```diff
@@ -6,219 +6,278 @@
...
+    } else if (respond_to_request(responder, request->neighbor, request->hash,
+                                  &pack) != RC_OK) {
+      log_warning(RESPONDER_LOGGER_ID, "Replying to request failed\n");
     }
+    transaction_request_queue_pop(&responder->queue);
```
th0br0: If we don't have the requested tx in our DB we should add it to our own TX req queue I believe?

Either way: we can pop instead of peek here as we always pop? (and thus reducing the time we hold the lock)

th0br0 (changes requested):

gossip/components/responder.c
thibault-martinez: We do request missing tx [here](https://github.com/iotaledger/entangled/pull/484/files/ec15c796b1cb3c08d3cefaf4bea6b3abf02b5646#diff-6c871dd3004f1b74db143dffd9d4ffe4R101).

thibault-martinez: pop only deletes the first element, it doesn't return it, and peek returns a pointer to the first element. What we can do to reduce the lock holding time is `lock, peek, make a copy of the element, pop, unlock, proceed to function` ?

th0br0 (approved):

PR #487, opened 2018-11-07
gossip: tips FIFO cache

### LINKED ISSUE #486: Tips FIFO cache

thibault-martinez:
Fixes #486

# Test Plan:
Added tests


th0br0 (approved):

th0br0 (changes requested):
Also add a helper `fifo_pop` method?
Also, could we rename `add` to `push`?

th0br0 (commented):
Isn't it counterintuitive to have a FIFO queue backed by a set?

thibault-martinez:
The main motivation for a set was the efficiency of removal.
e.g. removing a tip from a 5000 tips cache.

gossip/tips_cache.c
```diff
@@ -0,0 +1,218 @@
...
+  if (hash243_set_size(set) >= capacity) {
+    if ((ret = hash243_set_remove_entry(set, *set)) != RC_OK) {
+      return ret;
+    }
+  }
+
+  return hash243_set_add(set, tip);
+}
+
+static retcode_t tips_cache_random_tip_from_set(hash243_set_t* const set,
```
th0br0, line 32: Move this outside? This is rather a generic `random_hash_from_hash243_set` method?

gossip/tips_cache.c
```diff
@@ -0,0 +1,218 @@
...
+
+retcode_t tips_cache_init(tips_cache_t* const cache, size_t const capacity) {
+  if (cache == NULL) {
+    return RC_NULL_PARAM;
+  }
+
+  cache->tips = NULL;
+  cache->solid_tips = NULL;
+  cache->capacity = capacity;
+  rw_lock_handle_init(&cache->lock);
```
th0br0: Might be worth adding a note here that we could split this into two distinct locks which might net us a performance improvement

gossip/tips_cache.c
```diff
@@ -0,0 +1,218 @@
...
+
+  if (cache == NULL || tip == NULL) {
+    return RC_NULL_PARAM;
+  }
+
+  rw_lock_handle_rdlock(&cache->lock);
+
+  if (hash243_set_size(&cache->solid_tips) == 0) {
+    rw_lock_handle_unlock(&cache->lock);
+    return tips_cache_random_tip(cache, tip);
```
th0br0: Calling this method here is counterintuitive? This also relocks the rdlock again

gossip/tips_cache.h
```diff
@@ -0,0 +1,135 @@
...
+
+#ifndef __GOSSIP_TIPS_H__
+#define __GOSSIP_TIPS_H__
+
+#include "common/errors.h"
+#include "common/trinary/flex_trit.h"
+#include "utils/containers/hash/hash243_set.h"
+#include "utils/handles/rw_lock.h"
+
+#define MAX_TIPS 5000
```
th0br0, line 16: not used, remove

th0br0 (changes requested):

gossip/tips_cache.h
thibault-martinez: It's being used on #491

th0br0: Fair enough; as this is just stored in the `capacity` member I'd much rather not have this in a global header but in the source file that initialises the tips_cache?

thibault-martinez: Sure. As it can affect the result of `getTips` API call, should it be configurable with an option from CLI ?

th0br0: :thinking: there's two distinct params here - how many tips are returned from API call and how many are stored in memory. I believe the two are tracked independently on IRI, not sure

(but ya, might be good)

thibault-martinez: Just looking at IRI conf code, I don't see any option to limit the number of tips returned by the call

gossip/tips_cache.c
thibault-martinez: I'll add the second lock as part of this PR

gossip/tips_cache.h
thibault-martinez: CLI option will be added in #491

gossip/tips_cache.c
thibault-martinez: Would you return null hash as well if there is no solid hash ?

th0br0 (approved):

PR #489, opened 2018-11-08
Get cIRI database path from configuration variables

### LINKED ISSUE #132: Get cIRI database path from configuration variables

wusyong:
Fixes #132 
- Add `db_path`

# Test Plan:
CI must pass


consensus/conf.bzl
```diff
@@ -1,4 +1,5 @@
 CONSENSUS_MAINNET_VARIABLES = [
+    "DB_PATH='\"ciri/ciri.db\"'",
```
thibault-martinez: That's really great, now that you're at it, can you make it `ciri/ciri-mainnet.db` ?

consensus/conf.bzl
```diff
@@ -7,6 +8,7 @@ CONSENSUS_MAINNET_VARIABLES = [
 ]
 
 CONSENSUS_TESTNET_VARIABLES = [
+    "DB_PATH='\"ciri/ciri.db\"'",
```
thibault-martinez: And this one `ciri/ciri-testnet.db` ?

ciri/usage.h
```diff
@@ -95,6 +96,7 @@ static struct cli_argument_s {
      "Limits how many milestones behind the current one the random walk can "
      "start.",
      REQUIRED_ARG},
+    {"db-path", CLI_ARG_DB_PATH, "Path to the database file.", REQUIRED_ARG},
```
thibault-martinez: Maybe add it a short option `-d` instead of `CLI_ARG_DB_PATH `

thibault-martinez (changes requested):

thibault-martinez:
@wusyong Thank you for this really great PR !
A general comment, could you make this option reside under the cIRI conf instead of the consensus conf ?

consensus/conf.bzl
wusyong: Should I also duplicate a database file for this one?

thibault-martinez: Yes, remove the old one, create the two new and update the paths.

```
rm ciri/ciri.db
mkdir ciri/db
sqlite3 ciri/db/ciri-mainnet.db < common/storage/sql/schema.sql
sqlite3 ciri/db/ciri-testnet.db < common/storage/sql/schema.sql
```


This is temporary as we want bazel to do that for us in the future.

ciri/usage.h
```diff
@@ -98,7 +99,7 @@ static struct cli_argument_s {
 
     {NULL, 0, NULL, NO_ARG}};
 
-static char* short_options = "hl:n:t:u:p:";
+static char* short_options = "hld:n:t:u:p:";
```
thibault-martinez: You are missing a `:` after `l`

thibault-martinez (changes requested):

thibault-martinez:
Why are the initial databases files 44k ? Are they really empty, containing only the schema ?
I think they should be 20480 bytes.

wusyong:
Sorry, didn't aware they were modified when testing.

thibault-martinez (approved):

thibault-martinez:
Thanks @wusyong this is very useful. Good first PR !

ciri/BUILD
```diff
@@ -1,10 +1,26 @@
...
+    name = "testnet",
+    values = {"define": "network=testnet"},
+)
+
 cc_binary(
     name = "ciri",
     srcs = ["main.c"],
     copts = ["-DLOGGER_ENABLE"],
-    data = ["ciri.db"],
+    data = [
```
th0br0: Please use a `select({":testnet": ["db/ciri-testnet.db"], ":mainnet": ["db/ciri-mainnet.db"]})` here, no need to package both simultaneously.

ciri/BUILD
```diff
@@ -1,10 +1,26 @@
...
+
+config_setting(
+    name = "mainnet",
+    values = {"define": "network=mainnet"},
+)
+
+config_setting(
+    name = "testnet",
+    values = {"define": "network=testnet"},
+)
```
th0br0: Do not create new  `config_setting` entries for this.
Instead, we should move the generic `config_setting` flags to, e.g., `//config/BUILD` and then reuse them here.
(e.g. right now, we already have `//gossip:mainnet` and `//gossip:testnet` that describe this parameter)

th0br0 (changes requested):
Just some nits for the buildfiles :)

ciri/BUILD
wusyong: `//consensus` also has same `config_setting` entries. Should I modify them all together?

th0br0: That'd be awesome :)

th0br0 (approved):

PR #492, opened 2018-11-08
cclient: add libcclient.so target

th0br0:
We need a cc_binary target to create a shared library that's linked all dependencies in.
This can then be exported and linked against from other targets.


th0br0:
Skipping CI because this is a pure devops change.

PR #506, opened 2018-11-12
cclient: fixing the threshold misunderstanding.

oopsmonk:
the meaning of threshold is different in get_inputs and get_balances.
refer to [createGetInputs.ts](https://github.com/iotaledger/iota.js/blob/next/packages/core/src/createGetInputs.ts) and [createGetBalances.ts](https://github.com/iotaledger/iota.js/blob/next/packages/core/src/createGetBalances.ts)

# Test Plan:
CI must pass


cclient/iota_client_extended_api.c
```diff
@@ -128,12 +128,15 @@ retcode_t iota_client_get_inputs(iota_client_service_t const* const serv,
   // get balances from all addresses
   get_balances_req_t* blances_req = get_balances_req_new();
   get_balances_res_t* blances_res = get_balances_res_new();
-  blances_req->threshold = threshold;
+  blances_req->threshold = 100;  // currently `100` should be used
```
th0br0: could you also address the `blances` typo?

cclient/response/get_balances.c
```diff
@@ -37,13 +37,18 @@ char* get_balances_res_balances_at(get_balances_res_t const* const in,
...
-size_t get_balances_res_total_balance(get_balances_res_t const* const res) {
+size_t get_balances_res_total_balance(get_balances_res_t const* const res,
+                                      size_t const threshold) {
   size_t sum = 0;
-  size_t elm_count = utarray_len(res->balances);
-  for (int i = 0; i < elm_count; i++) {
-    char** p = (char**)utarray_eltptr(res->balances, i);
-    char* endptr;
-    sum += strtol(*p, &endptr, 10);
+  size_t balance = 0;
```
th0br0: should use & return `uint64_t`

cclient/response/get_balances.c
```diff
@@ -37,13 +37,18 @@ char* get_balances_res_balances_at(get_balances_res_t const* const in,
...
+  char** p = NULL;
+  char* endptr;
+
+  while ((p = (char**)utarray_next(res->balances, p))) {
+    balance = strtol(*p, &endptr, 10);
```
th0br0: we move the string->int conversion to the JSON layer

th0br0 (changes requested):

oopsmonk:
@th0br0 thanks for your advice, typo and conversion are fixed.

cclient/iota_client_extended_api.c
```diff
@@ -126,18 +126,21 @@ retcode_t iota_client_get_inputs(iota_client_service_t const* const serv,
...
-  ret_code = iota_client_get_balances(serv, blances_req, blances_res);
-  out_input->total_balance = get_balances_res_total_balance(blances_res);
+  ret_code = iota_client_get_balances(serv, balances_req, balances_res);
 
-  blances_req->addresses = NULL;  // no need to be freed
-  get_balances_req_free(&blances_req);
-  get_balances_res_free(&blances_res);
+  // filter balances via threshold
+  out_input->total_balance =
+      get_balances_res_total_balance(balances_res, threshold);
```
wusyong: According to [filterByThreshold](https://github.com/iotaledger/iota.js/blob/next/packages/core/src/createGetInputs.ts#L147), `threshold` in `get_inputs` should mean "total balance >= threshold" instead of "each balance > threshold". And input addresses should also be filtered I think.

oopsmonk: yes, you are right, thanks.

th0br0 (approved):

PR #513, opened 2018-11-13
mam/v2: integration - standardize types

tsvisabo:
The purpose of this PR is to integrate MAM-V2 with entangled's standard types

- Add some methods to trit_array to allow needed flexibility
- Use trit_t/flex_trits whenever possible
- fix some missing initializations 
- fix edge case bug


mam/v2/buffers.h
```diff
@@ -47,4 +45,4 @@ MAM2_API size_t buffers_copy_to(buffers_t *tb, trits_t buf);
  * copied. */
 MAM2_API size_t buffers_copy_from(buffers_t *tb, trits_t buf);
```
jserv, line 46: Should `buffers_copy_from` be removed?

tsvisabo: We don't know yet

Issue #253, opened 2018-09-06
Implement storeAndBroadcast


oopsmonk:

thibault-martinez:
Closed by #530

Issue #245, opened 2018-09-06
Implement findTransactionObjects


oopsmonk:

thibault-martinez:
Closed by #530

Issue #244, opened 2018-09-06
Implement getTransactionObjects


oopsmonk:

thibault-martinez:
Closed by #530

Issue #242, opened 2018-09-06
Implement getLatestInclusion


oopsmonk:

thibault-martinez:
Closed by #530

PR #530, opened 2018-11-15
cclient: implementing find_transaction_objects.

oopsmonk:
Fixing issue #245 #244 #248 #242 #253 #255 

# Test Plan:
CI must pass


cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,63 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+    transaction_array_t out_tx_objs) {
+  retcode_t ret_code = RC_OK;
+  get_trytes_res_t* out_trytes = get_trytes_res_new();
+  hash8019_queue_entry_t* q_iter = NULL;
+  iota_transaction_t tx = NULL;
+
+  ret_code = iota_client_get_trytes(serv, tx_hashes, out_trytes);
+  if (ret_code == RC_OK) {
+    CDL_FOREACH(out_trytes->trytes, q_iter) {
+      tx = transaction_deserialize(q_iter->hash);
```
thibault-martinez: Maybe use `transaction_serialize_on_flex_trits` on a preallocated `flex_trit_t [FLEX_TRIT_SIZE_8019]` to avoid allocation ?

thibault-martinez (changes requested):

cclient/iota_client_extended_api.c
oopsmonk: you mean `transaction_deserialize_from_trits` right? yes, for avoiding dynamic allocation, I can use `struct _iota_transaction tx` instead of `transaction_deserialize`.

thibault-martinez: oops yes, that was what I meant, still waking up here :)

oopsmonk: never mind, thanks for your suggestion.

cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,236 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+        log_error(CCLIENT_EXTENDED_LOGGER_ID, "%s: %s.\n", __func__,
+                  error_2_string(ret_code));
+        goto done;
+      }
+    }
+  }
+
+done:
+  check_consistency_req_free(&consistency_req);
+  check_consistency_res_free(consistency_res);
```
thibault-martinez, line 365: Free `out_trytes` as well ?

cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,236 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+  check_consistency_res_t* consistency_res = check_consistency_res_new();
+  get_trytes_res_t* out_trytes = get_trytes_res_new();
+  struct _iota_transaction tx = {};
+  size_t tx_deserialize_offset = 0;
+  uint64_t time_now = 0;
+  // Check if transaction wasn't issued before last 6 milestones
+  // Milestones are being issued every ~2mins, and 1 minute delay for
+  // propagating the network of nodes.
+  uint64_t max_depth = (6 * 2 - 1) * (60 * 1000);
+  log_debug(CCLIENT_EXTENDED_LOGGER_ID, "[%s:%d]\n", __func__, __LINE__);
```
thibault-martinez: Empty log ?

cclient/serialization/json/json_serializer.c
```diff
@@ -490,6 +490,64 @@ static retcode_t hash8019_stack_to_json_array(hash8019_stack_t stack,
   return RC_OK;
 }
 
+static retcode_t hash8019_array_to_json_array(hash8019_array_p array,
+                                              cJSON* const json_root,
+                                              char const* const obj_name) {
+  size_t array_count = 0;
+  cJSON* array_obj = NULL;
+  trit_t trytes_out[NUM_TRYTES_SERIALIZED_TRANSACTION + 1];
```
thibault-martinez: Should be `tryte_t`

thibault-martinez: And do we really need the `+ 1` here ?

utils/containers/hash/tests/test_hash_array.c
```diff
@@ -0,0 +1,53 @@
...
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "utils/containers/hash/hash_array.h"
+#include "utils/containers/hash/tests/defs.h"
+
+void test_empty_hash() {
+  flex_trit_t null_hash[FLEX_TRIT_SIZE_243] = {[0 ... FLEX_TRIT_SIZE_243 - 1] =
```
thibault-martinez: I think this is a gcc-only extension 🤔

utils/containers/hash/hash_array.c
```diff
@@ -0,0 +1,66 @@
+/*
```
thibault-martinez, line 1: Can we make this file on the same pattern than the others ? i.e. using bazel template ?

thibault-martinez (changes requested):

cclient/serialization/json/json_serializer.c
oopsmonk: The ending char is required by cJSON_CreateString.

utils/containers/hash/hash_array.c
oopsmonk: It's no need to use the template since the implementation is quite simple, only the `hashxxx_array_new` are separated.

th0br0 (commented):
Not sure what to think of the hash_array.
This will probably lead to issues because people accidentally push flex_trits of the wrong length into a statically sized array?

oopsmonk:
The same problem with hash_queue, hash_stack as well, there is no way to check if people pushed a flex_trits pointer with a wrong length, and we agreed not to use `trit_array_t` in cclient. 
The `hash_array` takes less memory and can do random access, but problems are 1. do we need random access? 2. the performance of expanding array size.

utils/containers/hash/hash_array.c
```diff
@@ -0,0 +1,66 @@
...
+
+flex_trit_t* hash_array_at(hash_array_p array, size_t index) {
+  return (flex_trit_t*)utarray_eltptr(array, index);
+}
+
+size_t hash_array_len(hash_array_p array) { return utarray_len(array); }
+
+void hash_array_free(hash_array_p array) { utarray_free(array); }
+
+bool empty_hash(flex_trit_t const* const hash, size_t len) {
```
ifullgaz: Maybe this function should be in flex_trit? It can be useful elsewhere

cclient/types/types.c
```diff
@@ -170,3 +170,27 @@ retcode_t flex_hash_to_char_buffer(trit_array_p hash, char_buffer_t* out) {
   }
   return RC_OK;
 }
+
+static UT_icd ut_transactions_icd = {sizeof(struct _iota_transaction), NULL,
```
ifullgaz, line 174: Shouldn't transaction_free be in the dtor member of the icd?

th0br0: utarray already calls free

utils/containers/hash/hash_array.c
```diff
@@ -0,0 +1,66 @@
...
+  return array;
+}
+
+hash8019_array_p hash8019_array_new() {
+  hash8019_array_p array = NULL;
+  utarray_new(array, &ut_hash8019_icd);
+  return array;
+}
+
+void hash_array_push(hash_array_p array, flex_trit_t const* const trits) {
```
th0br0, line 46: Given our discussion just now on Slack, could you add a `hash_array_reserve` method here?

utils/containers/hash/hash_array.c
th0br0: Renaming this to sth more flex_trit specific sounds best to me as well

utils/containers/hash/hash_array.h
```diff
@@ -0,0 +1,47 @@
...
+#include "utarray.h"
+
+#include "common/trinary/flex_trit.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef UT_array* hash_array_p;
+typedef UT_array* hash27_array_p;
```
th0br0: let's typedef the hash27... instances as `typedef hash_array_p hash27_array_p`?

th0br0 (changes requested):

utils/containers/hash/hash_array.c
oopsmonk: done

common/trinary/flex_trit.c
```diff
@@ -385,3 +385,13 @@ size_t flex_trits_from_bytes(flex_trit_t *to_flex_trits, size_t to_len,
 #endif
   return num_trits;
 }
+
+bool is_empty_flex_trits(flex_trit_t const *const hash, uint32_t len) {
```
thibault-martinez: We already have this function (`flex_trits_are_null`)

common/trinary/flex_trit.h
```diff
@@ -278,6 +278,9 @@ size_t flex_trits_to_bytes(byte_t *bytes, size_t to_len,
 size_t flex_trits_from_bytes(flex_trit_t *to_flex_trits, size_t to_len,
                              const byte_t *bytes, size_t len, size_t num_trits);
 
+// return true if all flex_trit are NULL, else false.
+bool is_empty_flex_trits(flex_trit_t const *const hash, uint32_t len);
```
thibault-martinez: We already have this function (`flex_trits_are_null`)

utils/containers/hash/tests/test_hash_array.c
```diff
@@ -0,0 +1,53 @@
...
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "utils/containers/hash/hash_array.h"
+#include "utils/containers/hash/tests/defs.h"
+
+void test_empty_hash() {
```
thibault-martinez: This test should be in `flex_trit_t` tests ?

utils/containers/hash/hash_array.h
```diff
@@ -0,0 +1,46 @@
...
+typedef hash_array_p hash27_array_p;
+typedef hash_array_p hash81_array_p;
+typedef hash_array_p hash243_array_p;
+typedef hash_array_p hash6561_array_p;
+typedef hash_array_p hash8019_array_p;
+
+hash27_array_p hash27_array_new();
+hash81_array_p hash81_array_new();
+hash243_array_p hash243_array_new();
+hash243_array_p hash6561_array_new();
```
thibault-martinez: `hash6561_array_p`

utils/containers/hash/hash_array.h
```diff
@@ -0,0 +1,46 @@
...
+hash243_array_p hash8019_array_new();
```
thibault-martinez: `hash8019_array_p`

utils/containers/hash/hash_array.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#ifndef __UTILS_CONTAINERS_HASH_HASH243_ARRAY_H__
```
thibault-martinez: Can we remove `243` here ?

utils/containers/hash/hash_array.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "utarray.h"
+
+#include "common/trinary/flex_trit.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef UT_array* hash_array_p;
```
thibault-martinez, line 19: Can we instead do `typedef UT_array* hash_array_t;` (like you've done with `typedef UT_array* transaction_array_t;`) ? Because here the base type is actually a pointer. I find `hash_array_p` to be a bit misleading because we can think we have a pointer to the type when it's not really the case.

thibault-martinez (changes requested):

utils/containers/hash/hash_array.h
oopsmonk: `UT_array` is a structure, not a pointer. so `hash_array_p` is a pointer to the UT_array structure.

thibault-martinez: Yes, what I meant to say was that the base type we are going to handle is a `UT_array *`. It makes sense since an array IS a pointer.

thibault-martinez: Anyway, one of `hash_array_p` or `transaction_array_t` has to change since they are contradictory, and IMHO it should be `hash_array_p`.

oopsmonk: IMO, `hash_array_p` and `transaction_array_t` (should be `transaction_array_p`) are two different object, hash_array is a hash container and transaction_array holds `struct _iota_transaction`.

oopsmonk: let's change `transaction_array_t` to `transaction_array_p`.

thibault-martinez: They are both arrays containing different kind of objects.
As you want, but IMHO it would make more sense to have `_t` because even in C what we call an array is just a pointer so the type itself IS a pointer.

oopsmonk: fair enough.

oopsmonk: In this case, we should change `trit_array_p` to `trit_array_t` as well, right?

thibault-martinez: That is a good question and I actually think that this `trit_array_p` typedef shouldn't even exists (for reasons mentioned here https://github.com/iotaledger/entangled/pull/589#discussion_r238860288) and we are in a very similar situation. So why not just `typedef UT_array hash_array_t;` and use `hash_array_t *` when needed ?

thibault-martinez: Anyway, the PR looks good to me I'll approve.
I was just sharing thoughts, I find names like `hash_array_p` very confusing and I think we should stay away from typedefing pointers.

thibault-martinez (approved):

utils/containers/hash/hash_array.h
oopsmonk: It's about code readability, we agree not to use `_t` as the suffix of a structure pointer and `_p` is a reminder for me that it's a pointer, b/c I am too lazy to add `*`.  **`¯\_(⌣̯̀⌣́)_/¯`**
agree, to use typedefs of non-pointer types only.

th0br0 (approved):

PR #538, opened 2018-11-17
storage: fix index creation in SQL schema

wusyong:
`hash_index` on `iota_milestone` wasn't created actually because of duplicate index name.

# Test Plan:
CI must pass


thibault-martinez:
Hi @wusyong, thank your  for this PR, you're actually right !

Most SQL documentations say
> index_name Is the name of the index. Index names must be unique within a table or view but do not have to be unique within a database. Index names must follow the rules of identifiers.

But SQLite also says
> If the optional IF NOT EXISTS clause is present and another index with the same name already exists, then this command becomes a no-op.

thibault-martinez (changes requested):
`ciri/db/ciri-mainnet.db` and `ciri/db/ciri-testnet.db` still contain the old `CREATE INDEX hash_index ON iota_transaction(hash);`, could you remove it from them please ?

thibault-martinez (approved):

Issue #238, opened 2018-09-06
Implement broadcastBundle 


oopsmonk:

oopsmonk:
fixed in #540

PR #540, opened 2018-11-19
cclient/extended_api: iota_client_get_bundle and iota_client_traverse_bundle

ifullgaz:
iota_client_get_bundle and iota_client_traverse_bundle extended APIs

# Test Plan:
Usual test plan


cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,80 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+    tx = transaction_deserialize(tmp_trytes);
+    if (!tx) {
+      ret_code = RC_CCLIENT_OOM;
+      goto cleanup;
+    }
+    get_trytes_res_free(&get_trytes_res);
+    // Check that the first transaction we get is really a tail transaction
+    // Its index must be 0 and the list of transactions empty.
+    // It is an error if the first transaction has index > 0
+    if (!utarray_len(tx_objs) && transaction_current_index(tx)) {
```
th0br0: `tx_objs` is not a utarray

cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,80 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+  get_trytes_res_t* get_trytes_res = NULL;
+  iota_transaction_t tx;
+  flex_trit_t* tmp_trytes;
+  flex_trit_t* trunk_hash = (flex_trit_t *)trunk_tx;
+  
+  log_debug(CCLIENT_EXTENDED_LOGGER_ID, "[%s:%d]\n", __func__, __LINE__);
+
+  do {
+    // Create request and response to get_trytes API
+    get_trytes_req = get_trytes_req_new();
```
th0br0: We should reuse these instead of reallocing?

th0br0 (changes requested):
Also add a test case?

cclient/iota_client_extended_api.c
ifullgaz: tx_objs is a transaction_objs_t `typedef UT_array transaction_objs_t`.
Should it be another type?

cclient/iota_client_extended_api.c
ifullgaz: Ah, yes

cclient/iota_client_extended_api.c
th0br0: sorry, I overlooked your changed definition.
You'll still have to initialise it via `utarray_new` and provide an ICD though?
https://troydhanson.github.io/uthash/utarray.html

ifullgaz: Yes, I thought that the caller would do that, unless you think it's more appropriate to do it in the function, in which case I'll change tx_objs to `transaction_objs_t **tx_objs`?

cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,80 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+    }
+    get_trytes_req_free(&get_trytes_req);
+    // Get the transaction trytes
+    tmp_trytes = hash8019_queue_at(&get_trytes_res->trytes, 0);
+    if (!tmp_trytes) {
+      ret_code = RC_CCLIENT_RES_ERROR;
+      goto cleanup;
+    }
+    // Create a new transaction with the received trytes
+    tx = transaction_deserialize(tmp_trytes);
```
thibault-martinez: `transaction_deserialize` does a dynamic allocation. No need for that here ? Use a statically allocated transaction object + `transaction_deserialize_from_trits`

cclient/iota_client_extended_api.c
```diff
@@ -249,3 +249,80 @@ retcode_t iota_client_get_account_data(iota_client_service_t const* const serv,
...
+    }
+    get_trytes_res_free(&get_trytes_res);
+    // Check that the first transaction we get is really a tail transaction
+    // Its index must be 0 and the list of transactions empty.
+    // It is an error if the first transaction has index > 0
+    if (!utarray_len(tx_objs) && transaction_current_index(tx)) {
+      ret_code = RC_CCLIENT_INVALID_TAIL_HASH;
+      goto cleanup;
+    }
+    // If the transaction is valid, add it to the transactino list
```
thibault-martinez: Typo

cclient/iota_client_extended_api.h
```diff
@@ -21,9 +21,9 @@ typedef flex_hash_array_t address_list_t;
 typedef flex_hash_array_t inputs_list_t;
 typedef flex_hash_array_t transfer_list_t;
 typedef flex_hash_array_t transaction_list_t;
-typedef flex_hash_array_t transaction_objs_t;
 typedef flex_hash_array_t booleans_t;
 typedef flex_hash_array_t input_list_t;
+typedef UT_array transaction_objs_t;
```
thibault-martinez: We already have `typedef UT_array bundle_transactions_t;` and related methods in `common/model/bundle.h`. Would it make sense, for semantic, to use them ?

thibault-martinez (changes requested):

cclient/iota_client_extended_api.h
oopsmonk: Does `bundle_transactions_t` equal to a list of transaction object? IMO, it has different meaning between a bundle and a transactions list.

thibault-martinez: `bundle_transactions_t` is an array of `iota_transaction_t`, the exact same thing as `transaction_objs_t`. But since here we are dealing with bundles maybe we should use it ?

oopsmonk: ah..right, it is a bundle operation. sorry.

cclient/iota_client_extended_api.c
ifullgaz: We potentially want to return several transactions

cclient/iota_client_extended_api.c
ifullgaz: Ok

cclient/iota_client_extended_api.c
thibault-martinez: But `tx` will be copied into the array right ? So no need to have it dynamic ?

oopsmonk:
marked as not merger due to APIs still WIP, including `iota_client_get_bundle`, `iota_client_promote_transaction`, `iota_client_replay_bundle`

oopsmonk:
TODO `iota_client_promote_transaction`

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,148 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
   return ret_code;
 }
 
+static retcode_t check_consistency(iota_client_service_t const* const serv,
```
th0br0: this method should support multiple tails

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,148 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+    log_warning(logger_id, "check consistency failed\n");
+    goto cleanup;
+  }
+  if (!consistency_res->state) {
+    ret_code = RC_CCLIENT_CONSISTENCY;
+  }
+
+cleanup:
+  check_consistency_req_free(&consistency_req);
+  check_consistency_res_free(consistency_res);
```
th0br0: missing a &?

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,148 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+    log_error(logger_id, "%s check_consistency_res_new failed: %s\n", __func__,
+              error_2_string(ret_code));
+    goto cleanup;
+  }
+
+  hash243_queue_push(&consistency_req->tails, tail_tx);
+  ret_code =
+      iota_client_check_consistency(serv, consistency_req, consistency_res);
+  if (ret_code) {
+    log_warning(logger_id, "check consistency failed\n");
```
th0br0: also log ret_code?

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,148 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+      ret_code = RC_CCLIENT_RES_ERROR;
+      log_error(logger_id, "%s hash8019_queue_at failed: %s\n", __func__,
+                error_2_string(ret_code));
+      goto cleanup;
+    }
+    // Create a new transaction with the received trytes
+    transaction_deserialize_from_trits(&tx, tmp_trytes, true);
+
+    current_index = transaction_current_index(&tx);
+    last_index = transaction_last_index(&tx);
```
th0br0: we should only read the last_index once (for first tx in bundle)

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,148 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+    if (!tmp_trytes) {
+      ret_code = RC_CCLIENT_RES_ERROR;
+      log_error(logger_id, "%s hash8019_queue_at failed: %s\n", __func__,
+                error_2_string(ret_code));
+      goto cleanup;
+    }
+    // Create a new transaction with the received trytes
+    transaction_deserialize_from_trits(&tx, tmp_trytes, true);
+
+    current_index = transaction_current_index(&tx);
```
th0br0: should probably make sure that we start at current_index = 0?

th0br0 (changes requested):

cclient/iota_client_extended_api.c
oopsmonk: check tail at line#167

cclient/iota_client_extended_api.c
oopsmonk: considering the logic here, it won't cause a problem and keep the code simple.

oopsmonk:
fixing #240, #252, #238 
let's do `iota_client_promote_transaction` in another PR.

common/model/transaction.c
```diff
@@ -351,6 +352,28 @@ size_t transaction_deserialize_from_trits(iota_transaction_t *const transaction,
...
+  memset(message, '9', sizeof(message));
+  flex_trits_from_trytes(transaction_message(spam_transaction),
+                         NUM_TRITS_SIGNATURE, message, NUM_TRITS_SIGNATURE,
+                         NUM_TRYTES_SIGNATURE);
+  flex_trits_from_trytes(transaction_address(spam_transaction),
+                         NUM_TRITS_ADDRESS, message, NUM_TRITS_ADDRESS,
+                         NUM_TRYTES_ADDRESS);
+  flex_trits_from_trytes(transaction_tag(spam_transaction), NUM_TRITS_TAG,
+                         message, NUM_TRITS_TAG, NUM_TRYTES_TAG);
+  return spam_transaction;
```
thibault-martinez: `transaction_new()` already sets these fields to `FLEX_TRIT_NULL_VALUE`

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,108 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+  return ret_code;
+
+cleanup:
+  if (bundle) {
+    utarray_clear(bundle);
+  }
+  if (trytes) {
+    hash_array_free(trytes);
+  }
+  // done:
```
thibault-martinez: remove ?

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,108 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+      goto cleanup;
+    }
+    // Call the get_trytes API
+    ret_code = iota_client_get_trytes(serv, get_trytes_req, get_trytes_res);
+    if (ret_code != RC_OK) {
+      log_error(logger_id, "%s iota_client_get_trytes failed: %s\n", __func__,
+                error_2_string(ret_code));
+      goto cleanup;
+    }
+    hash243_queue_free(&get_trytes_req->hashes);
```
thibault-martinez: `hash243_queue_pop` would achieve the same with less variables

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,108 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+    // Get the transaction trytes
+    tmp_trytes = hash8019_queue_at(&get_trytes_res->trytes, 0);
```
thibault-martinez: Same here with `hash8019_queue_peek`

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,108 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+    if (!tmp_trytes) {
+      ret_code = RC_CCLIENT_RES_ERROR;
+      log_error(logger_id, "%s hash8019_queue_at failed: %s\n", __func__,
+                error_2_string(ret_code));
+      goto cleanup;
+    }
+    // Create a new transaction with the received trytes
+    transaction_deserialize_from_trits(&tx, tmp_trytes, true);
```
thibault-martinez: We can avoid hashing since we already have the hash, we just need to set it

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,108 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+
+    current_index = transaction_current_index(&tx);
+    last_index = transaction_last_index(&tx);
+    // Get the next tail_hash from the transaction
+    memcpy(trunk_hash, transaction_trunk(&tx), FLEX_TRIT_SIZE_243);
```
thibault-martinez: Can't `trunk_hash` just be a pointer to avoid copies ?

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,108 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+        ret_code = RC_CCLIENT_INVALID_TAIL_HASH;
+        log_error(logger_id, "%s not a tail transaction: %s\n", __func__,
+                  error_2_string(ret_code));
+        goto cleanup;
+      }
+      // The transaction is valid, add it to the transaction list
+      bundle_transactions_add(bundle, &tx);
+    }
+    // The response is not needed anymore
+    hash8019_queue_free(&get_trytes_res->trytes);
```
thibault-martinez, line 158: `hash8019_queue_pop` too

cclient/iota_client_extended_api.c
```diff
@@ -546,3 +645,148 @@ retcode_t iota_client_send_transfer(
...
+
+  broadcast_transactions_req = broadcast_transactions_req_new();
+  if (!broadcast_transactions_req) {
+    ret_code = RC_CCLIENT_NULL_PTR;
+    log_error(logger_id, "%s broadcast_transactions_req_new failed: %s\n",
+              __func__, error_2_string(ret_code));
+    goto cleanup;
+  }
+
+  log_debug(logger_id, "[%s:%d]\n", __func__, __LINE__);
```
thibault-martinez: What are these for ? Can we remove them ?

cclient/iota_client_extended_api.c
```diff
@@ -546,3 +645,148 @@ retcode_t iota_client_send_transfer(
...
+
+bundle_status_t iota_client_get_bundle(iota_client_service_t const* const serv,
+                                       flex_trit_t const* const tail_hash,
+                                       bundle_transactions_t* const bundle) {
+  bundle_status_t bundle_status = BUNDLE_NOT_INITIALIZED;
+  log_debug(logger_id, "[%s:%d]\n", __func__, __LINE__);
+  if (iota_client_traverse_bundle(serv, tail_hash, bundle) == RC_OK) {
+    bundle_validator(bundle, &bundle_status);
+  }
+  return bundle_status;
```
thibault-martinez: Should this be a parameter ? We are dismissing the real error

cclient/iota_client_extended_api.c
```diff
@@ -546,3 +645,148 @@ retcode_t iota_client_send_transfer(
...
+      iota_client_check_consistency(serv, consistency_req, consistency_res);
+  if (ret_code != RC_OK) {
+    log_error(logger_id, "%s checking consistency failed: %s\n", __func__,
+              error_2_string(ret_code));
+    goto done;
+  }
+
+  if (consistency_res->state) {
+    // adding spam transaction
+    spam_transaction = transaction_new();
```
thibault-martinez, line 752: Just noting for the next PR that this need to be freed

thibault-martinez (changes requested):

common/model/transaction.c
oopsmonk: missed this method, let's remove it.

cclient/iota_client_extended_api.c
oopsmonk: right, thanks.

cclient/iota_client_extended_api.c
oopsmonk: I don't think so.

cclient/serialization/tests/test_json.c
```diff
@@ -584,21 +589,23 @@ void test_serialize_attach_to_tangle(void) {
       trits_8019, NUM_TRITS_SERIALIZED_TRANSACTION,
       (const tryte_t*)TEST_2673_TRYTES_1, NUM_TRYTES_SERIALIZED_TRANSACTION,
       NUM_TRYTES_SERIALIZED_TRANSACTION));
-  hash_array_push(attach_req->trytes, trits_8019);
+  hash_array_push(trytes, trits_8019);
```
thibault-martinez, line 592: I really think we should offer API functions like `attach_to_tangle_req_add_trytes` for this kind of operations. The end user is not supposed to know the underlying types and their functions.

thibault-martinez (changes requested):

cclient/serialization/tests/test_json.c
oopsmonk: good point, see issue #805

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,107 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+  get_trytes_res = get_trytes_res_new();
+  if (!get_trytes_res) {
+    ret_code = RC_CCLIENT_OOM;
+    log_error(logger_id, "%s get_trytes_res_new failed: %s\n", __func__,
+              error_2_string(ret_code));
+    goto cleanup;
+  }
+
+  do {
+    memcpy(trunk_hash, is_tail ? tail_hash : transaction_trunk(&tx),
```
th0br0: we don't need the local `trunk_hash` here, can simply call `transaction_set_trunk(tx, tail_hash)` before the loop and always refer to the hash inside the loop via `transaction_trunk`

cclient/iota_client_extended_api.c
```diff
@@ -56,6 +56,107 @@ static retcode_t is_unused_address(iota_client_service_t const* const serv,
...
+      // save raw trytes if we want them.
+      hash_array_push(trytes, tmp_trytes);
+    }
+    if (bundle) {
+      // add this transaction to bundle
+      bundle_transactions_add(bundle, &tx);
+    }
+    // The response is not needed anymore
+    hash8019_queue_free(&get_trytes_res->trytes);
+  } while (current_index != last_index);
```
th0br0, line 160: Doesn't the JS implementation do additional checks here?
What if the bundle we're trying to traverse is malformed maliciously? e.g. varying indices, ...
Ideally, we should at least check that the `current_index` of the retrieved transaction matches our expected index (`current_index+1`) and maybe also check that bundle hash is consistent

th0br0 (changes requested):

cclient/iota_client_extended_api.c
oopsmonk: yes, JS doesn't have additional checks.
I expect the current_index is in the right order(0,1,2..) and the bundle hash is consistent since transactions are fetched from the Tangle.

cclient/iota_client_extended_api.c
oopsmonk: `trunk_hash` log the bundle hash for next transaction that we don't recalculate it during deserialization.

th0br0: no, it holds the tail tx's hash (i.e. H(tx in bundle with curIdx=0), ...)

cclient/iota_client_extended_api.c
th0br0: getTrytes doesn't do any server-side tx validation iirc

cclient/iota_client_extended_api.c
oopsmonk: done.

cclient/iota_client_extended_api.c
oopsmonk: done.

th0br0 (approved):

thibault-martinez (approved):

PR #559, opened 2018-11-23
conf: add a coordinator CLI option

### LINKED ISSUE #461: Add a coordinator option
Add a CLI/file option to set the coordinator address, similar to IRI's `--testnet-coordinator`.

thibault-martinez:
Fixes #461 

# Test Plan:
CI must pass


th0br0 (changes requested):
What's this needed for? We're reading the coordinator from the snapshot json file, no?

thibault-martinez:
We do, but with the snapshot json file we are limited to either the mainnet coo or the testnet coo but not a custom one. This is the equivalent of IRI's `--testnet-coordinator`

th0br0:
Right, but people could just as well provide an alternative snapshot.json file? 
That'd be better IMHO because they might also want to change the starting index as part of passing in a separate coordinator address?

th0br0 (approved):

PR #587, opened 2018-12-03
common: refactoring transfer model.

### LINKED ISSUE #578: common: add error codes and helper methods in the transfer model.

oopsmonk:
This PR fixes #578.
* refactoring and adding helper functions and error handling in the transfer model.
* converting DOS to UNIX format in `common/model/transfer.c` file.

# Test Plan:
CI must pass.

### DIFF
--- a/common/model/transfer.c
+++ b/common/model/transfer.c
@@ -349,7 +349,8 @@ void transfer_ctx_hash(transfer_ctx_t* transfer_ctx, Kerl* kerl,
 }
 
 transfer_iterator_t* transfer_iterator_new(transfer_t* transfers[], size_t len,
-                                           Kerl* kerl) {
+                                           Kerl* kerl,
+                                           iota_transaction_t transaction) {
   transfer_ctx_t* transfer_ctx = NULL;
   transfer_iterator_t* transfer_iterator =
       (transfer_iterator_t*)calloc(1, sizeof(transfer_iterator_t));
@@ -358,6 +359,16 @@ transfer_iterator_t* transfer_iterator_new(transfer_t* transfers[], size_t len,
               __LINE__);
     return NULL;
   }
+
+  if (transaction) {
+    transfer_iterator->transaction = transaction;
+    transfer_iterator->dynamic_transaction = 0;
+  } else {
+    // Dynamically allocate a transaction structure
+    transfer_iterator->transaction = transaction_new();
+    transfer_iterator->dynamic_transaction = 1;
+  }
+
   transfer_iterator->transfers = transfers;
   transfer_iterator->transfers_count = len;
   transfer_ctx = transfer_ctx_new();
@@ -401,12 +412,6 @@ iota_transaction_t transfer_iterator_next(
 
   if (transfer_iterator->current_transfer <
       transfer_iterator->transfers_count) {
-    // Dynamically allocate a transaction structure if none was
-    // previously provided with transfer_iterator_set_transaction
-    if (!transfer_iterator->transaction) {
-      transfer_iterator->transaction = transaction_new();
-      transfer_iterator->dynamic_transaction = 1;
-    }
     transfer_t* transfer =
         transfer_iterator->transfers[transfer_iterator->current_transfer];
     size_t count = transfer_transactions_count(transfer);
--- a/common/model/transfer.c
+++ b/common/model/transfer.c
@@ -7,7 +7,7 @@
 
 #include "common/model/transfer.h"
 
-#define TRANSFER_LOGGER_ID "cclient_core_api"
+#define TRANSFER_LOGGER_ID "transfer"
 
 /***********************************************************************************************************
  * Private interface
@@ -37,7 +37,7 @@ static void transfer_iterator_next_data_transaction(
                   __func__, __LINE__);
     }
   } else {
-    log_error(TRANSFER_LOGGER_ID, "[%s:%d] transfer type not match.\n",
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] the transfer type doesn't match.\n",
               __func__, __LINE__);
   }
 }
@@ -75,7 +75,7 @@ static void transfer_iterator_next_output_transaction(
                   __func__, __LINE__);
     }
   } else {
-    log_error(TRANSFER_LOGGER_ID, "[%s:%d] transfer type not match.\n",
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] the transfer type doesn't match.\n",
               __func__, __LINE__);
   }
 }
@@ -87,10 +87,10 @@ static void transfer_iterator_next_input_transaction(
 
   if (transfer->type == VALUE_IN) {
     value_in = (transfer_value_in_t*)transfer->meta;
-    memcpy(tx->signature_or_message, value_in->data, sizeof(value_in->len));
+    memcpy(tx->signature_or_message, value_in->data, value_in->len);
     tx->value = transfer->value;
   } else {
-    log_error(TRANSFER_LOGGER_ID, "[%s:%d] transfer type not match.\n",
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] the transfer type doesn't match.\n",
               __func__, __LINE__);
   }
 }
@@ -272,7 +272,7 @@ transfer_t* transfer_value_in_new(flex_trit_t const* const address,
 }
 
 void transfer_free(transfer_t** transfer) {
-  if (transfer || *transfer) {
+  if (transfer && *transfer) {
     free((*transfer)->meta);
     free(*transfer);
     *transfer = NULL;
@@ -383,7 +383,7 @@ transfer_iterator_t* transfer_iterator_new(transfer_t* transfers[], size_t len,
 
 void transfer_iterator_free(transfer_iterator_t** iter) {
   transfer_iterator_t* iter_p = *iter;
-  if (*iter || iter) {
+  if (iter && *iter) {
     if (iter_p->dynamic_transaction && iter_p->transaction) {
       free(iter_p->transaction);
     }
--- a/common/model/transfer.c
+++ b/common/model/transfer.c
@@ -4,539 +4,401 @@
  *
  * Refer to the LICENSE file for licensing information
  */
-#include <string.h>
 
-#include "common/helpers/sign.h"
 #include "common/model/transfer.h"
-#include "common/trinary/trit_long.h"
 
-/***********************************************************************************************************
- * Transfer Input data structure
- ***********************************************************************************************************/
-// To obfuscate the model, the structure definition comes here
-// struct _transfer_output {...};
-
-// Get the seed
-const flex_trit_t *transfer_output_seed(transfer_output_t transfer_output) {
-  return transfer_output->seed;
-}
-
-// Set the seed (assign)
-void transfer_output_set_seed(transfer_output_t transfer_output,
-                              const flex_trit_t *seed) {
-  transfer_output->seed = seed;
-}
-
-// Get the security level
-size_t transfer_output_security(transfer_output_t transfer_output) {
-  return transfer_output->security;
-}
-
-// Set the security level
-void transfer_output_set_security(transfer_output_t transfer_output,
-                                  size_t security) {
-  transfer_output->security = security;
-}
-
-// Get the index
-size_t transfer_output_seed_index(transfer_output_t transfer_output) {
-  return transfer_output->seed_index;
-}
-
-// Set the index
-void transfer_output_set_seed_index(transfer_output_t transfer_output,
-                                    size_t index) {
-  transfer_output->seed_index = index;
-}
-
-// Creates and returns a new transfer output
-transfer_output_t transfer_output_new(void) {
-  transfer_output_t transfer_output;
-  transfer_output = (transfer_output_t)malloc(sizeof(struct _transfer_output));
-  if (!transfer_output) {
-    // errno = IOTA_OUT_OF_MEMORY
-    return NULL;
-  }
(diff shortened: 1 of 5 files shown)


common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+    }
+    transaction = transfer_iterator->transaction;
+    // Reset all transaction fields
+    transaction_reset(transaction);
+    // Set common transaction fields
+    memcpy(transaction->bundle, transfer_iterator->bundle_hash,
+           sizeof(transaction->bundle));
+    memcpy(transaction->address, transfer->address,
+           sizeof(transaction->address));
+    memcpy(transaction->obsolete_tag, transfer->tag,
```
thibault-martinez: Shouldn't it be `transfer->obsolete_tag` ?

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+    len = len > NUM_TRITS_SIGNATURE ? NUM_TRITS_SIGNATURE : len;
+    flex_ret =
+        flex_trits_slice(transfer_iterator->transaction->signature_or_message,
+                         len, trans_data->data, data_len, offset, len);
+    if (flex_ret == 0) {
+      log_warning(TRANSFER_LOGGER_ID, "[%s:%d] flex_trits slicing failed.\n",
+                  __func__, __LINE__);
+    }
+  } else {
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] transfer type not match.\n",
```
tsvisabo: do not

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+        NUM_TRITS_SIGNATURE * output_info->security,
+        NUM_TRITS_SIGNATURE *
+            transfer_iterator->current_transfer_transaction_index,
+        NUM_TRITS_SIGNATURE);
+    if (flex_ret == 0) {
+      log_warning(TRANSFER_LOGGER_ID, "[%s:%d] flex_trits slicing failed.\n",
+                  __func__, __LINE__);
+    }
+  } else {
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] transfer type not match.\n",
```
tsvisabo: do not

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+
+// create data transfer object.
+transfer_t* transfer_data_new(flex_trit_t const* const address,
+                              flex_trit_t const* const tag,
+                              flex_trit_t const* const data, size_t data_len,
+                              uint64_t timestamp) {
+  transfer_data_t* tf_data = NULL;
+  transfer_t* transfer = NULL;
+  if (!address) {
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] address cannot be NULL.\n", __func__,
```
tsvisabo, line 135: i think "must not" is preferable

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+
+// create a value_out transfer object.
+transfer_t* transfer_value_out_new(transfer_value_out_t const* const output,
+                                   flex_trit_t const* const tag,
+                                   flex_trit_t const* const address,
+                                   int64_t value, uint64_t timestamp) {
+  transfer_value_out_t* value_out = NULL;
+  transfer_t* tf = NULL;
+  if (!validate_output(output)) {
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] output is invalid.\n", __func__,
```
tsvisabo, line 183: cannot -> must not

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+              __LINE__);
+  }
+  transfer_ctx_free(transfer_ctx);
+  transfer_iterator->iota_signature_gen = iota_flex_sign_signature_gen;
+  return transfer_iterator;
+}
+
+void transfer_iterator_free(transfer_iterator_t** iter) {
+  transfer_iterator_t* iter_p = *iter;
+  if (*iter || iter) {
```
tsvisabo: Did you mean `&&`, also, you need to verify `iter` is not null before dereferencing

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+    }
+    transfer_t* transfer =
+        transfer_iterator->transfers[transfer_iterator->current_transfer];
+    size_t count = transfer_transactions_count(transfer);
+    if (transfer_iterator->current_transfer_transaction_index >= count) {
+      transfer_iterator->current_transfer++;
+      transfer_iterator->current_transfer_transaction_index = 0;
+      return transfer_iterator_next(transfer_iterator);
+    }
+    transaction = transfer_iterator->transaction;
```
tsvisabo, line 423: could you export that logic to a separate function?

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+
+iota_transaction_t transfer_iterator_next(
+    transfer_iterator_t* transfer_iterator) {
+  iota_transaction_t transaction = NULL;
+
+  if (transfer_iterator->current_transfer <
+      transfer_iterator->transfers_count) {
+    // Dynamically allocate a transaction structure if none was
+    // previously provided with transfer_iterator_set_transaction
+    if (!transfer_iterator->transaction) {
```
tsvisabo: doesn't this method suppose to be a getter/read only?
should we not return null in that case?

tsvisabo (changes requested):

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include "common/model/transfer.h"
+
+#define TRANSFER_LOGGER_ID "cclient_core_api"
```
thibault-martinez: Wrong logger id

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+}
+
+static void transfer_iterator_next_input_transaction(
+    transfer_iterator_t* transfer_iterator, transfer_t* transfer) {
+  transfer_value_in_t* value_in = NULL;
+  iota_transaction_t tx = transfer_iterator->transaction;
+
+  if (transfer->type == VALUE_IN) {
+    value_in = (transfer_value_in_t*)transfer->meta;
+    memcpy(tx->signature_or_message, value_in->data, sizeof(value_in->len));
```
thibault-martinez: `value_in->len` instead of `sizeof(value_in->len)` ?

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+  if (output->security < 1 || output->security > 3) {
+    log_error(TRANSFER_LOGGER_ID, "[%s:%d] security leve error.\n", __func__,
+              __LINE__);
+    return false;
+  }
+  return true;
+}
+
+// create data transfer object.
+transfer_t* transfer_data_new(flex_trit_t const* const address,
```
thibault-martinez, line 128: What about `_init` functions that takes and fills a struct instead of `_new` functions that allocs and fills a struct ?

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+  tf->address = address;
+  tf->tag = tag;
+  tf->value = value;
+  tf->timestamp = timestamp;
+  tf->meta = (void*)value_in;
+  return tf;
+}
+
+void transfer_free(transfer_t** transfer) {
+  if (transfer || *transfer) {
```
thibault-martinez: `&&`

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+      value_out = (transfer_value_out_t*)tf->meta;
+      return value_out->security;
+    case VALUE_IN:
+      // fixed
+      return 1;
+  }
+  return 0;
+}
+
+transfer_ctx_t* transfer_ctx_new() {
```
thibault-martinez, line 300: Do we need dynamic allocation for the ctx ?

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+  memset(tf_ctx->bundle, FLEX_TRIT_NULL_VALUE, FLEX_TRIT_SIZE_243);
+  return tf_ctx;
+}
+
+void transfer_ctx_free(transfer_ctx_t* transfer_ctx) { free(transfer_ctx); }
+
+bool transfer_ctx_init(transfer_ctx_t* transfer_ctx, transfer_t* transfers[],
+                       size_t len) {
+  size_t i;
+  int64_t total = 0;
```
thibault-martinez, line 316: Maybe set `transfer_ctx->count = 0` before the loop in case we re-init it ?

common/model/transfer.c
```diff
@@ -1,676 +1,474 @@
...
+  tf->meta = (void*)value_in;
+  return tf;
+}
+
+void transfer_free(transfer_t** transfer) {
+  if (transfer || *transfer) {
+    free((*transfer)->meta);
+    free(*transfer);
+    *transfer = NULL;
+  }
```
thibault-martinez, line 279: We should probably wipe out the `seed` field if the transfer is a `VALUE_OUT`

thibault-martinez (changes requested):

common/model/transfer.c
oopsmonk: it's no difference in this case.

common/model/transfer.c
```diff
@@ -272,7 +272,7 @@ transfer_t* transfer_value_in_new(flex_trit_t const* const address,
 }
 
 void transfer_free(transfer_t** transfer) {
-  if (transfer || *transfer) {
+  if (*transfer) {
```
thibault-martinez: We still need to check `transfer` before dereferencing it

common/model/transfer.c
```diff
@@ -383,7 +383,7 @@ transfer_iterator_t* transfer_iterator_new(transfer_t* transfers[], size_t len,
 
 void transfer_iterator_free(transfer_iterator_t** iter) {
   transfer_iterator_t* iter_p = *iter;
-  if (*iter || iter) {
+  if (*iter) {
```
thibault-martinez: We still need to check `iter` before dereferencing it

thibault-martinez (changes requested):

common/model/transfer.c
oopsmonk: that is a pointer in `transfer_value_out_t`

common/model/transfer.c
oopsmonk: should not re-init an init method.

common/model/transfer.c
oopsmonk: not sure, but just keep the logic in the previous change.

common/model/transfer.c
oopsmonk: it allocates transfer_t.

thibault-martinez: Yes, but why ? Is this really necessary ?
By taking a `transfer_t *` as parameter we allow the user to fill a static or a dynamic `transfer_t`. IMHO we should avoid dynamic allocation as much as possible in code used in the cclient.

common/model/transfer.c
thibault-martinez: Fair enough

common/model/transfer.c
oopsmonk: what do you mean? could you give more details?

common/model/transfer.c
oopsmonk: this change just provides helper methods and keeps original design in this model. 
In my understanding, structures are mostly pointers in the transfer.

common/model/transfer.c
oopsmonk: I think this logic is here only.

common/model/transfer.c
tsvisabo: you are allocating a new transaction, should we do that when someone calls `next`?

oopsmonk: it depends on the user, the comment on line 404 has answered your question.

common/model/transfer.c
thibault-martinez: Yes but since it will only be used by the client maybe it's a good thing to adapt it to the client needs. Pointers can be addresses of static structures. We'll probably need to remove dynamic allocation someday.

common/model/transfer.c
thibault-martinez: I think we don't need it

common/model/transfer.c
thibault-martinez: Maybe put a comment then ?

thibault-martinez (approved):

common/model/transfer.c
tsvisabo: yeah, it is not crucial, but it makes the code clearer

oopsmonk: Why do you think it's not clear? I mean no complex logic and it has some comments here.

common/model/transfer.c
oopsmonk: btw, why it exists?

thibault-martinez: I have no idea :)

oopsmonk: then, why do you think we don't need it?

thibault-martinez: Looking where it's being used (e.g. `transfer_iterator_new`), I don't understand why it's being dynamically allocated

oopsmonk: I prefer not to change something doesn't fully understand, we can seek for the answer from @th0br0 and @ifullgaz.

th0br0: Feel free to remove the method, it was an experiment. But in hindsight, I'm not that happy with the pointer aliases either.

tsvisabo (approved):

PR #589, opened 2018-12-04
Model separation

tsvisabo:
- all metadata fields view model
- snapshot index view model
- solid state view model
- all viewmodels require branch + trunk so it branching is possible
- need to figure out a way to do conditionals inside `#define` so we can have `DECLARE_PACK_SINGLE_GEN` that base on provided `load_model_t` arg decides which specific macro to call and generate `iota_stor_pack`


common/model/transaction.h
```diff
@@ -94,6 +103,52 @@ struct _iota_transaction {
   bool solid;
```
thibault-martinez: Should we remove metadata from the original `struct _iota_transaction` then ?

common/model/transaction.h
```diff
@@ -54,6 +54,15 @@ extern "C" {
  * Transaction data structure
  ***********************************************************************************************************/
 typedef struct _iota_transaction *iota_transaction_t;
+typedef struct _iota_transaction_meta_view_model
+    *iota_transaction_meta_view_model_t;
```
thibault-martinez: (I'm being super picky here)

Can we just do `typedef struct iota_transaction_meta_view_model_s iota_transaction_meta_view_model_t` and not involve pointer ? 

Having a typedefed pointer prevents us from being able to apply constness on the pointed data e.g.:

With typedef:
`void fn(iota_transaction_meta_view_model_t const vm)` (const pointer)

Without typedef:
`void fn(iota_transaction_meta_view_model_t const * const vm)` (const data const pointer)
or
`void fn(iota_transaction_meta_view_model_t * const vm)` (const pointer)

I know `typedef struct _iota_transaction *iota_transaction_t;` was already like that but I've been wanting to fix that for a long time :)

common/model/transaction.h
```diff
@@ -94,6 +103,52 @@ struct _iota_transaction {
   bool solid;
 };
 
+struct _iota_transaction_meta_view_model {
```
thibault-martinez, line 100: - Why not a `struct _iota_transaction` + metadata fields ?
- When do we use a `struct _iota_transaction_meta_view_model` instead of a `struct _iota_transaction` ?

common/storage/sql/statements.h
```diff
@@ -32,6 +32,10 @@ extern char* iota_statement_transaction_exist;
 extern char* iota_statement_transaction_exist_by_hash;
 extern char* iota_statement_transaction_approvers_count;
 
+extern char* iota_statement_transaction_select_meta_fields_by_hash;
+extern char* iota_statement_select_snapshot_by_hash;
```
thibault-martinez: Can you prefix `iota_statement_transaction_` ?

common/storage/sql/statements.h
```diff
@@ -32,6 +32,10 @@ extern char* iota_statement_transaction_exist;
...
+extern char* iota_statement_select_solid_by_hash;
```
thibault-martinez: Can you prefix `iota_statement_transaction_` ?

common/storage/storage.h
```diff
@@ -40,6 +40,15 @@ typedef enum transaction_field_e {
   TRANSACTION_FIELD_ADDRESS
 } transaction_field_t;
 
+typedef enum load_model_e {
+  MODEL_TRANSACTION,
+  MODEL_HASH,
+  MODEL_MILESTONE,
+  MODEL_TRANSACTION_META_ALL,
+  MODEL_SNAPSHOT_INDEX,
```
thibault-martinez, line 48: Can you prefix `MODEL_TRANSACTION_` ?

common/storage/storage.h
```diff
@@ -40,6 +40,15 @@ typedef enum transaction_field_e {
...
+  MODEL_SOLID,
```
thibault-martinez, line 49: Can you prefix `MODEL_TRANSACTION_` ?

thibault-martinez (changes requested):

common/model/transaction.h
thibault-martinez: https://github.com/iotaledger/entangled/issues/592

common/storage/sql/statements.h
tsvisabo: sure

common/model/transaction.h
tsvisabo: `struct _iota_transaction` should include all fields as sometimes they are all needed (e.g: in `bundle_validator`)

common/model/transaction.h
tsvisabo: will remove `sapshot_index` and `solid` from `struct _iota_transaction`

common/model/transaction.h
tsvisabo: of course

common/model/transaction.h
th0br0: I really think that this `view model` naming from IRI is terrible. Let's call these models?
Also, the models don't have links to which tx they're actually referring to?

th0br0 (changes requested):

common/model/transaction.h
tsvisabo: you mean they should also contain the `hash`?

th0br0: at the cost of using more memory, why not have a parent struct that contains the views as materialised entries  + a bitflag that tells which of these were loaded?

tsvisabo:
reopened as #598

th0br0:
Closing in favour of #598

PR #593, opened 2018-12-05
gossip: transaction requester thread

### LINKED ISSUE #590: Implement a transaction requester worker
Transaction requests are currently bound to incoming new packets from the gossip protocol.
The ratio amount of tx request / amount of new packets is not enough to keep up with the amount of requests.
The size of the request queue is always at its maximum (10000 being the default value).

thibault-martinez:
Fixes #590 

# Test Plan:
CI

### DIFF
--- a/gossip/components/transaction_requester.c
+++ b/gossip/components/transaction_requester.c
@@ -10,19 +10,44 @@
 #include "consensus/tangle/tangle.h"
 #include "utils/handles/rand.h"
 #include "utils/logger_helper.h"
+#include "utils/time.h"
 
 #define REQUESTER_LOGGER_ID "requester"
+#define REQUESTER_INTERVAL 100
+
+/*
+ * Private functions
+ */
+
+static void *transaction_requester_routine(
+    transaction_requester_t *const transaction_requester) {
+  if (transaction_requester == NULL) {
+    return NULL;
+  }
+
+  while (transaction_requester->running) {
+    fprintf(stderr, "Requester\n");
+    sleep_ms(REQUESTER_INTERVAL);
+  }
+
+  return NULL;
+}
+
+/*
+ * Public functions
+ */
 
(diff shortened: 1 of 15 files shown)


gossip/components/transaction_requester.c
```diff
@@ -8,189 +8,298 @@
...
+                      .num_bytes = FLEX_TRIT_SIZE_243,
+                      .dynamic = 0};
+  DECLARE_PACK_SINGLE_TX(tx, txp, pack);
+
+  if (transaction_requester == NULL) {
+    return NULL;
+  }
+
+  while (transaction_requester->running) {
+    if (requester_size(transaction_requester) > REQUESTER_THRESHOLD) {
```
th0br0: let's use a label here and invert the condition?

th0br0 (changes requested):

gossip/components/transaction_requester.c
```diff
@@ -8,189 +8,298 @@
...
+        if (ret == RC_OK && pack.num_loaded != 0) {
+          transaction_serialize_on_flex_trits(txp, transaction);
+        } else {
+          memset(transaction, FLEX_TRIT_NULL_VALUE, FLEX_TRIT_SIZE_8019);
+        }
+      }
+      if (transaction_requester->node->neighbors) {
+        iter = transaction_requester->node->neighbors->front;
+        while (iter) {
+          if (neighbor_send(transaction_requester->node, &iter->data,
```
th0br0: is this thread safe?

thibault-martinez: Not 100% thread-safe, I planned on refactoring the whole neighbour layer when doing `getNeighbors`, `addNeighbors` and `removeNeighbors` API calls (maybe next)

thibault-martinez: Fixed in #599

th0br0 (approved):

th0br0 (approved):

PR #598, opened 2018-12-05
Model separation 

tsvisabo:
#589 replaces
- all metadata fields view model
- snapshot index view model
- solid state view model
- all models require branch + trunk so it branching is possible
- removes metadata fields from iota_transaction


common/storage/sql/schema.sql
```diff
@@ -15,7 +15,7 @@ CREATE TABLE IF NOT EXISTS iota_transaction (
   attachment_timestamp_upper TIMESTAMP NOT NULL,
   nonce BLOB NOT NULL,
   hash BLOB NOT NULL PRIMARY KEY,
-  snapshot_index INTEGER NOT NULL,
+  snapshot_index INTEGER NOT NULL DEFAULT 0,
```
thibault-martinez, line 18: Can you reflect that on `ciri-mainnet.db` and `ciri-testnet.db` ?

consensus/milestone_tracker/milestone_tracker.c
```diff
@@ -173,8 +173,9 @@ static void* milestone_validator(void* arg) {
       hash243_queue_pop(&mt->candidates);
       rw_lock_handle_unlock(&mt->candidates_lock);
       hash_pack_reset(&pack);
-      if (iota_tangle_transaction_load(mt->tangle, TRANSACTION_FIELD_HASH,
-                                       &hash, &pack) == RC_OK &&
+      if (iota_tangle_transaction_selective_load(
+              mt->tangle, TRANSACTION_FIELD_HASH, &hash, &pack,
+              MODEL_TRANSACTION_META_ALL) == RC_OK &&
```
thibault-martinez, line 178: Note to myself: we actually only need the obsolete tag from this transaction so we might want to load only this field

thibault-martinez (changes requested):

tsvisabo:
as we are taking a new model separation, this PR is not relevant

Issue #153, opened 2018-08-15
Processor component enhancements


thibault-martinez:
- [x] Cache recent seen bytes
- [x] Adding non-tethered neighbors on Testnet
- [x] Processor thread pool
- [x] Validate transaction (#144)
- [x] Store transaction metadata (#80)

PR #599, opened 2018-12-06
ciri: neighbors thread safety

thibault-martinez:
Ensure neighbours thread safety for the incoming `getNeighbors `, `addNeighbors `and `removeNeighbors ` API calls.

# Test Plan:
Ci must pass


gossip/components/processor.c
```diff
@@ -198,35 +196,47 @@ static retcode_t process_packet(processor_t const *const processor,
...
         RC_OK) {
       log_warning(PROCESSOR_LOGGER_ID, "Processing request bytes failed\n");
-      return ret;
+      goto done;
     }
   } else {
+    log_debug(PROCESSOR_LOGGER_ID,
+              "Discarding packet from non-tethered node %s://%s:%d\n", protocol,
+              packet->source.ip, packet->source.port);
     // TODO Testnet add non-tethered neighbor
```
th0br0, line 234: do we have an issue for this?

gossip/neighbor.h
```diff
@@ -81,6 +83,72 @@ retcode_t neighbor_send_packet(node_t *const node, neighbor_t *const neighbor,
 retcode_t neighbor_send(node_t *const node, neighbor_t *const neighbor,
                         flex_trit_t const *const transaction);
 
+/**
+ * Adds a neighbor to a neighbors list
```
th0br0, line 87: Add comments explaining that the caller must hold the neighbours lock?

gossip/components/processor.c
```diff
@@ -198,35 +196,47 @@ static retcode_t process_packet(processor_t const *const processor,
...
     log_debug(PROCESSOR_LOGGER_ID, "Processing request bytes\n");
     if ((ret = process_request_bytes(processor, neighbor, packet, hash)) !=
         RC_OK) {
       log_warning(PROCESSOR_LOGGER_ID, "Processing request bytes failed\n");
-      return ret;
+      goto done;
     }
   } else {
+    log_debug(PROCESSOR_LOGGER_ID,
+              "Discarding packet from non-tethered node %s://%s:%d\n", protocol,
```
th0br0, line 232: This should have some feedback via return codes to close the connection?

th0br0 (changes requested):

gossip/components/processor.c
thibault-martinez: https://github.com/iotaledger/entangled/issues/153

gossip/components/processor.c
thibault-martinez: TCP connections from non-tethered nodes are automatically closed from within the `tcp_receiver` so this is mostly for incoming UDP packet from non-tethered nodes.
Though, it can happen that a previously paired TCP neighbor is unpaired with the API call without the connection being closed, but returning from `process_packet` won't allow the receiver to close the connection since it doesn't have access to it.

th0br0 (approved):

PR #610, opened 2018-12-11
gossip: replace iota_packet_queue implementation with liblfds queue

th0br0:
Swapped out utlist with libfds queue

# Test Plan:
Nonexistant tests pass

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -4,7 +4,7 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository", "new_git_r
 
 git_repository(
     name = "rules_iota",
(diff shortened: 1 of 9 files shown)


th0br0:
Something to note:
the _count method is singlethreaded only and we should actually lock around it...
or just risk an endless loop ? http://www.liblfds.org/mediawiki/index.php?title=r7.1.1:Function_lfds711_queue_umm_query

th0br0:
Just saw a segv with this, investigating.

th0br0:
Just to document the segv:
queue elements need to be reused, so we need to add a freelist here that keeps track of allocated but currently unused elements.

th0br0:
Superceded by #663

PR #611, opened 2018-12-11
common/model - partial loading

tsvisabo:
- Add bitmask on transaction struct for asserting loaded fields
- implement partial loading for solidity check and tangle traversals
- missing specialized partial loading for random walk and milestone validation


common/model/transaction.c
```diff
@@ -123,6 +123,7 @@ size_t transaction_deserialize_trits(iota_transaction_t transaction,
   curl_digest(tx_trits, NUM_TRITS_SERIALIZED_TRANSACTION, hash, &curl);
   flex_trits_from_trits(transaction->consensus.hash, NUM_TRITS_HASH, hash,
                         NUM_TRITS_HASH, NUM_TRITS_HASH);
+  transaction->loaded_columns_mask |= MASK_ALL_COLUMNS;
```
thibault-martinez: Is it really all columns if solid and snapshot are not set ?

common/model/transaction.h
```diff
@@ -119,8 +120,34 @@ struct _iota_transaction {
   iota_transaction_fields_consensus_t consensus;
   iota_transaction_fields_data_t data;
   iota_transaction_fields_metadata_t metadata;
+  int64_t loaded_columns_mask;
```
thibault-martinez: Does it need to be 64 ? Also, unsigned ?

common/model/transfer.c
```diff
@@ -424,14 +424,11 @@ iota_transaction_t transfer_iterator_next(
...
-           sizeof(transaction_address(transaction)));
-    memcpy(transaction_obsolete_tag(transaction), transfer->tag,
-           sizeof(transaction_obsolete_tag(transaction)));
-    memcpy(transaction_tag(transaction), transfer->tag,
-           sizeof(transaction_tag(transaction)));
+    memcpy(transaction->essence.bundle, transfer_iterator->bundle_hash,
+           FLEX_TRIT_SIZE_243);
+    memcpy(transaction->essence.address, transfer->address, FLEX_TRIT_SIZE_243);
+    memcpy(transaction->essence.obsolete_tag, transfer->tag, FLEX_TRIT_SIZE_81);
+    memcpy(transaction->attachment.tag, transfer->tag, FLEX_TRIT_SIZE_81);
```
thibault-martinez: Why aren't we using the transaction setters here ?

consensus/ledger_validator/ledger_validator.c
```diff
@@ -261,8 +261,8 @@ retcode_t iota_consensus_ledger_validator_update_snapshot(
                                  .dynamic = 0};
   *has_snapshot = false;
 
-  if ((ret = iota_tangle_transaction_load(lv->tangle, TRANSACTION_FIELD_HASH,
-                                          &milestone_hash, &pack)) != RC_OK) {
+  if ((ret = iota_tangle_transaction_load_for_traversal(
+           lv->tangle, milestone->hash, &pack)) != RC_OK) {
```
thibault-martinez: We only need snapshot + solid here, it would make sense to only load metadata

thibault-martinez (changes requested):

common/model/transaction.h
tsvisabo: it does not have to be unsigned if you are only looking at bits

consensus/ledger_validator/ledger_validator.c
tsvisabo: you will need branch+trunk

thibault-martinez: Do I ? Looks like we are only using `transaction_solid` and `transaction_snapshot_index` in this function ?

tsvisabo: you are correct, confused it with traversal

common/model/transaction.c
thibault-martinez: I'm still wondering if `MASK_ALL_COLUMNS` is adequate since it's missing some fields 🤔

gossip/components/processor.c
```diff
@@ -40,7 +40,8 @@ static retcode_t process_transaction_bytes(processor_t const *const processor,
   retcode_t ret = RC_OK;
   bool exists = false;
   struct _iota_transaction transaction = {.metadata.snapshot_index = 0,
-                                          .metadata.solid = 0};
+                                          .metadata.solid = 0,
+                                          .loaded_columns_mask = 0};
```
thibault-martinez, line 44: Random thought: if we were to add some more fields to be defaulted, or if you think 3 are already forgettable, maybe we should add a `DECLARE_TX` macro or something

consensus/milestone_tracker/milestone_tracker.c
```diff
@@ -176,8 +172,8 @@ static void* milestone_validator(void* arg) {
       hash243_queue_pop(&mt->candidates);
       rw_lock_handle_unlock(&mt->candidates_lock);
       hash_pack_reset(&pack);
-      if (iota_tangle_transaction_load(mt->tangle, TRANSACTION_FIELD_HASH,
-                                       &hash, &pack) == RC_OK &&
+      if (iota_tangle_transaction_load_essence_and_consensus(
```
thibault-martinez: I think we only need the essence here

common/storage/sql/sqlite3/storage.c
```diff
@@ -315,25 +345,43 @@ static retcode_t execute_statement_load_transactions(
...
+static retcode_t execute_statement_load_transaction_essence_and_consensus(
+    sqlite3_stmt* const sqlite_statement, iota_stor_pack_t* const pack) {
+  return execute_statement_load_gen(sqlite_statement, pack, pack->capacity,
+                                    MODEL_TRANSACTION_MODEL_ESSENCE_CONSENSUS);
+}
+
+static retcode_t execute_statement_load_transaction_metadata(
+    sqlite3_stmt* const sqlite_statement, iota_stor_pack_t* const pack) {
+  return execute_statement_load_gen(sqlite_statement, pack, pack->capacity,
+                                    MODEL_TRANSACTION_MODEL_METADATA);
```
thibault-martinez, line 365: Do we need them ? Why not call `execute_statement_load_gen` directly ?

thibault-martinez (changes requested):

gossip/components/processor.c
tsvisabo: we could, but then that's more explicit IMO

thibault-martinez: Alright

common/storage/sql/schema.sql
```diff
@@ -15,7 +15,7 @@ CREATE TABLE IF NOT EXISTS iota_transaction (
   attachment_timestamp_upper TIMESTAMP NOT NULL,
   nonce BLOB NOT NULL,
   hash BLOB NOT NULL PRIMARY KEY,
-  snapshot_index INTEGER NOT NULL,
+  snapshot_index INTEGER NOT NULL DEFAULT 0,
```
thibault-martinez, line 18: Can you reflect this on `ciri/db/ciri-mainnet.db` as well ?

thibault-martinez (changes requested):

thibault-martinez (approved):

PR #619, opened 2018-12-14
api: find_transactions API call

### LINKED ISSUE #427: findTransactions implementation

thibault-martinez:
Fixes #427 

# Test Plan:
Added tests


ciri/api/tests/test_find_transactions.c
```diff
@@ -8,8 +8,103 @@
...
+static iota_consensus_t consensus;
+
+// Compares two queues by using an intermediary hash set
+static bool hash243_queue_cmp(hash243_queue_t const lhs,
+                              hash243_queue_t const rhs) {
+  hash243_set_t set = NULL;
+  hash243_queue_entry_t *iter = NULL;
+  bool cmp = true;
+
+  if (hash243_queue_count(lhs) != hash243_queue_count(rhs)) {
```
tsvisabo, line 28: What if a queue contains duplication/s?

common/storage/sql/sqlite3/storage.c
```diff
@@ -950,6 +950,139 @@ retcode_t iota_stor_transaction_approvers_count(connection_t const* const conn,
...
+  size_t approvees_count = hash243_queue_count(approvees);
+  hash243_queue_entry_t* iter243 = NULL;
+  hash81_queue_entry_t* iter81 = NULL;
+  size_t offset = 0;
+  size_t column = 1;
+  bool and = false;
+
+  // Base size of the query is 109 chars
+  // We then add space for the "?,"
+  char* statement = calloc(1, 109 + 2 * bundles_count + 2 * addresses_count +
```
tsvisabo: can we have a #define for 109 size?

tsvisabo: the logic of the dynamic statement creation should be in statements.c or another file

tsvisabo (changes requested):

common/storage/sql/sqlite3/storage.c
```diff
@@ -950,14 +950,44 @@ retcode_t iota_stor_transaction_approvers_count(connection_t const* const conn,
...
+    in_clause[offset - 1] = '\0';
+  }
+  return in_clause;
+}
+
+static char* iota_stor_transaction_find_build_query(
+    size_t const bundles_count, size_t const addresses_count,
+    size_t const tags_count, size_t const approvees_count) {
+  // Base size of the query + enough space for '?' (bindings)
+  size_t statement_size = 328 + 2 * bundles_count + 2 * addresses_count +
```
tsvisabo: This function build a statement and should be in statement.c

tsvisabo (changes requested):

ciri/api/tests/test_find_transactions.c
thibault-martinez: Then the test will fail, It is just a utility function used for this test case, we are comparing the resulting queue from the API call with a queue we construct ourselves so if they are not identical it won't pass.

common/storage/sql/sqlite3/storage.c
thibault-martinez: Sure

common/storage/sql/sqlite3/storage.c
thibault-martinez: Sure

ciri/api/tests/test_find_transactions.c
tsvisabo: ok, but this function is not in the test only namespace, and even though a queue should not contain dups, it should be tested when comparing because queue is not a set

thibault-martinez: I'm actually using an intermediary set in this function

thibault-martinez: What did you mean by "this function is not in the test only namespace" ? It actually is

tsvisabo (approved):

ciri/api/tests/test_find_transactions.c
tsvisabo: you're right

PR #631, opened 2018-12-19
api: check_consistency API call

### LINKED ISSUE #438: checkConsistency implementation

thibault-martinez:
Fixes #438 

# Test Plan:
Added tests


ciri/api/api.c
```diff
@@ -434,15 +434,72 @@ retcode_t iota_api_were_addresses_spent_from(
...
 
-  return RC_OK;
+  CDL_FOREACH(req->tails, iter) {
+    bundle_transactions_new(&bundle);
+    hash_pack_reset(&pack);
+    if ((ret = iota_tangle_transaction_load(&api->consensus->tangle,
+                                            TRANSACTION_FIELD_HASH, iter->hash,
+                                            &pack)) != RC_OK) {
+    } else if (pack.num_loaded == 0) {
+      ret = RC_API_TAIL_MISSING;
```
tsvisabo, line 461: why won't you set that in res?

tsvisabo (changes requested):

ciri/api/api.c
thibault-martinez: For IRI-compatibility, some errors are returned with an error response, some are included in the `info` field of the `checkConsistency` response...

thibault-martinez: Because some are errors due to non-respect of the API contract, some are just that the tails are not consistent and `info` tells why

tsvisabo (approved):

PR #632, opened 2018-12-19
common/model: arrival_time

tsvisabo:
- Add arrival as metadata column which is indexed in db


th0br0 (approved):

thibault-martinez (commented):
Can you actually set it when new txs are received ? Or another PR ?

th0br0:
It's set automatically (see the `DEFAULT CURRENT_TIMESTAMP`)

PR #644, opened 2018-12-28
storage: per-thread tangle connection

thibault-martinez:
This PR switches sqlite3 to multi-threaded mode by implementing per-thread tangle connection.
Per-thread prepared statements will be added in another PR.

# Test Plan:
CI

### DIFF
--- a/ciri/conf.c
+++ b/ciri/conf.c
@@ -78,7 +78,10 @@ static retcode_t set_conf_value(iota_ciri_conf_t* const ciri_conf,
   switch (key) {
     // cIRI configuration
     case 'd':  // --db-path
-      strcpy(ciri_conf->db_path, value);
+      strncpy(ciri_conf->db_path, value, sizeof(ciri_conf->db_path));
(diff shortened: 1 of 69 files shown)


ciri/conf.c
```diff
@@ -207,7 +210,11 @@ retcode_t iota_ciri_conf_default(iota_ciri_conf_t* const ciri_conf,
   }
 
   ciri_conf->log_level = DEFAULT_LOG_LEVEL;
-  strcpy(ciri_conf->db_path, DEFAULT_DB_PATH);
+  strncpy(ciri_conf->db_path, DEFAULT_DB_PATH, sizeof(ciri_conf->db_path));
```
th0br0, line 213: Let's use memcpy - strncpy will add \0 thus potentially blocking out the last character?

th0br0 (changes requested):
lgtm
Could we add a convention to always have `tangle` as the second or first arg to a method thougH?

th0br0 (approved):

PR #645, opened 2018-12-28
consensus/exit_probablility_randomizer: exit_prob_map

tsvisabo:
- On both virtual and non virtual functions inside `exit_prob_map`, accept explicitly the derived class ptr to 
  avoid casting


thibault-martinez (commented):
Can you have a look at all the warnings generated on the buildkite consoles ?

th0br0:
ping?

th0br0:
Tsvi Sabo [1:19 PM]
I think we can close it and I will reopen it as part of overall warnings PR

PR #647, opened 2018-12-29
solidifier: fixes and optimizations

thibault-martinez:
- Fixes an issue where transactions could be marked solid even if their trunk wasn't
- Removes a leak
- Optimises solidity propagation by switching sets pointers instead of allocating
- Reduces duration of locks

# Test Plan:
CI

### DIFF
--- a/consensus/transaction_solidifier/transaction_solidifier.c
+++ b/consensus/transaction_solidifier/transaction_solidifier.c
@@ -5,14 +5,11 @@
  * Refer to the LICENSE file for licensing information
  */
 
-#include "transaction_solidifier.h"
-
+#include "consensus/transaction_solidifier/transaction_solidifier.h"
 #include "consensus/utils/tangle_traversals.h"
-#include "utils/hash_indexed_map.h"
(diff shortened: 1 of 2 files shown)


consensus/transaction_solidifier/transaction_solidifier.c
```diff
@@ -292,25 +299,29 @@ static retcode_t check_transaction_and_update_solid_state(
...
         RC_OK) {
       *is_new_solid = false;
+      log_error(TRANSACTION_SOLIDIFIER_LOGGER_ID,
+                "Checking solidity of trunk failed\n");
       return ret;
     }
 
     if ((ret = check_approvee_solid_state(
-             ts, tangle, transaction_branch(transaction), is_new_solid)) !=
+             ts, tangle, transaction_branch(transaction), &is_branch_solid)) !=
```
tsvisabo, line 312: Instead of having to booleans, we should not query for branch solid state if trunk is not solid

tsvisabo (changes requested):

tsvisabo (approved):

PR #648, opened 2018-12-30
consensus/utils : generic key and value map

tsvisabo:


utils/containers/map.c.tpl
```diff
@@ -0,0 +1,66 @@
+
```
th0br0: remove whiteline

utils/containers/map.c.tpl
```diff
@@ -0,0 +1,66 @@
...
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include <search.h>
```
th0br0: not used?

th0br0 (changes requested):

th0br0 (approved):

PR #652, opened 2019-01-02
transaction_validator: various changes and documentations

thibault-martinez:
- Transaction timestamps as `uint64_t` instead of `int64_t`
- Transaction constness
- Transaction validator cleans
- Transaction validator documentations
- Use `flex_trits_at` instead of conversion to test last trit of addresses

# Test Plan:
Ci

### DIFF
--- a/consensus/transaction_validator/transaction_validator.c
+++ b/consensus/transaction_validator/transaction_validator.c
@@ -34,22 +34,24 @@ static uint64_t MAX_TIMESTAMP_FUTURE_MS = 2 * 60 * 60 * 1000;
  */
 static bool has_invalid_timestamp(transaction_validator_t const* const tv,
                                   iota_transaction_t const* const transaction) {
-  uint64_t max_future_timestamp_ms =
-      current_timestamp_ms() + MAX_TIMESTAMP_FUTURE_MS;
+  uint64_t timestamp_ms = transaction_attachment_timestamp(transaction) == 0
+                              ? transaction_timestamp(transaction) * 1000UL
+                              : transaction_attachment_timestamp(transaction);
+  bool is_too_futuristic =
+      timestamp_ms > (current_timestamp_ms() + MAX_TIMESTAMP_FUTURE_MS);
+  bool is_below_snapshot =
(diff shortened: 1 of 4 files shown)


consensus/transaction_validator/transaction_validator.c
```diff
@@ -14,67 +13,91 @@
...
   uint64_t max_future_timestamp_ms =
       current_timestamp_ms() + MAX_TIMESTAMP_FUTURE_MS;
+
   if (transaction_attachment_timestamp(transaction) == 0) {
     return (transaction_timestamp(transaction) <
                 tv->conf->snapshot_timestamp_sec &&
-            memcmp(tv->conf->genesis_hash, transaction_hash(transaction),
-                   FLEX_TRIT_SIZE_243)) ||
-           transaction_timestamp(transaction) * 1000 > max_future_timestamp_ms;
+            memcmp(transaction_hash(transaction), tv->conf->genesis_hash,
```
th0br0: if you're already editing this, could you pull out the individual conditions to local variables? just to clarify logical relationships

consensus/transaction_validator/transaction_validator.c
```diff
@@ -14,67 +13,91 @@
...
 }
 
 bool iota_consensus_transaction_validate(
-    transaction_validator_t* const tv, iota_transaction_t* const transaction) {
+    transaction_validator_t const* const tv,
+    iota_transaction_t const* const transaction) {
   if (transaction_weight_magnitude(transaction) < tv->conf->mwm) {
     log_debug(TRANSACTION_VALIDATOR_LOGGER_ID,
-              "Validation failed, invalid weight magnitude\n");
+              "Validation failed: insufficient PoW performed\n");
```
th0br0: the old text is better. let's call it `insufficient transaction weight`?
The amount of PoW you have to do varies, so saying that PoW performed is insufficient isn't really correct.

th0br0 (changes requested):
Yay - documentation! Thanks!

th0br0 (approved):

PR #655, opened 2019-01-04
gossip: tips solidifier implementation

### LINKED ISSUE #499: Implement tips solidifier
Current IRI logic is a thread that every 750ms:
- Randomly picks a non-solid tip from the cache
- Checks if it's still a tip
- If it is, tries to solidify it

Reference: https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/service/TipsSolidifier.java

thibault-martinez:
- Fixes #499
- Implements tips solidifier
- Fixes thread safety of `iota_consensus_transaction_solidifier_check_solidity` by removing `solid_transactions_candidates` from the `transaction_solidifier_t` struct and having it local to the function. It also allows to get rid of a lot of unnecessary locking.

# Test Plan:
CI


gossip/components/tips_solidifier.c
```diff
@@ -0,0 +1,172 @@
...
+
+    if (is_solid) {
+      if (tips_cache_set_solid(tips_solidifier->tips, tip) != RC_OK) {
+        log_warning(TIPS_SOLIDIFIER_LOGGER_ID,
+                    "Changing status of tip to solid failed\n");
+      }
+    }
+
+  sleep:
+    sleep_ms(TIPS_SOLIDIFICATION_INTERVAL_MS);
```
th0br0, line 89: Shouldn't we sleep less than 750ms if we've still got tips in the cache?

th0br0 (approved):

PR #657, opened 2019-01-05
consensus/solidifer : Avoid DFS when not needed

tsvisabo:
- rename set
- do not branch if current traversed hash is already found to be solid in previous DFS traversal


tsvisabo:
PR is not needed, overlooked 
`ret = iota_tangle_transactions_update_solid_state(
        tangle, solid_transactions_candidates, true);`

PR #660, opened 2019-01-06
common/trinary: trits and trytes conversions optimizations

thibault-martinez:
100000000 iterations
trits_to_trytes before  135222 ms
trits_to_trytes after   117607 ms (-13.027%)

100000000 iterations
trytes_to_trits before  187789 ms
trytes_to_trits after   113821 ms (-39.389%)

# Test Plan:
CI

### DIFF
--- a/common/trinary/trit_tryte.c
+++ b/common/trinary/trit_tryte.c
@@ -5,7 +5,6 @@
  * Refer to the LICENSE file for licensing information
  */
 
-#include <assert.h>
 #include <string.h>
 
 #include "common/defs.h"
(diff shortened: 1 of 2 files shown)


common/trinary/trit_tryte.c
```diff
@@ -68,34 +67,33 @@ uint8_t set_trit_at(tryte_t *const trytes, size_t const length, size_t index,
...
-    j += trits[i];
-  }
-  // int j = trits[0] + trits[1] * 3 + trits[2] * 9;
-  if (j < 0) {
-    j += TRYTE_SPACE;
-  }
-  trytes[0] = TRYTE_ALPHABET[(size_t)j];
-  if (length <= NUMBER_OF_TRITS_IN_A_TRYTE) {
-    return;
+  int k = 0;
```
tsvisabo, line 70: Cool, this function now supports trit arrays of more than just 3 trits, but what did the performance gain in this func?

tsvisabo (approved):

common/trinary/trit_tryte.c
thibault-martinez: It already supported trit arrays of more than just 3 trits but via recursion, switching from recursion to iteration had a some performance gain

th0br0: just a nit, but this should probably've been an `int8_t`

thibault-martinez: I actually tried that but performances were a little bit worse. I guess it's because CPUs are more efficient working on their word size ?

PR #664, opened 2019-01-07
utils: forced inline macro and inline sboxes

### LINKED ISSUE #608: Add macro to force inline methods
Inlining (on GCC) is often determined by the method size.
Some methods should probably always be inlined (e.g. transform, sbox in `common/curl/curl_p.c`).
We need a good compiler-specific macro for this.

thibault-martinez:
Fixes #608 

# Test Plan:
CI

### DIFF
--- a/common/curl-p/ptrit.h
+++ b/common/curl-p/ptrit.h
@@ -10,7 +10,6 @@
 
 #include "common/curl-p/const.h"
 #include "common/trinary/ptrit.h"
-#include "utils/forced_inline.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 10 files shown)


common/curl-p/trit.h
```diff
@@ -5,29 +5,39 @@
...
+extern "C" {
+#endif
 
 typedef struct {
   trit_t state[STATE_LENGTH];
   CurlType type;
 } Curl;
 
-void init_curl(Curl* ctx);
+static FORCED_INLINE void sbox(Curl* const c, Curl* const s) {
```
th0br0: move this to curl_p.c

common/curl-p/ptrit.h
```diff
@@ -9,26 +9,42 @@
...
 
 typedef struct {
   ptrit_t state[STATE_LENGTH];
   CurlType type;
 } PCurl;
 
-#ifdef __cplusplus
-extern "C" {
-#endif
+static FORCED_INLINE void ptrit_sbox(ptrit_t* const c, ptrit_t const* const s) {
```
th0br0: move this to `ptrit.c`

th0br0 (changes requested):

th0br0 (approved):

PR #665, opened 2019-01-08
common: implementing finalize_bundle.

### LINKED ISSUE #613: Model/bundle: About bundle finalization
Now we have `calculate_bundle_hash` to get bundle hash, 
I think a `bundle_finalize` function to fill in bundle hash through transactions is also needed.

Any idea?

oopsmonk:
implementing finalize_bundle and the test case. 

Fixes #613 

# Test Plan:
CI must pass

### DIFF
--- a/common/model/bundle.c
+++ b/common/model/bundle.c
@@ -11,6 +11,11 @@
 #include "common/trinary/tryte_long.h"
 
 static UT_icd bundle_transactions_icd = {sizeof(iota_transaction_t), 0, 0, 0};
+static const trit_t one[NUM_TRITS_TAG] = {
+    1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
 
 void bundle_transactions_new(bundle_transactions_t **const bundle) {
   utarray_new(*bundle, &bundle_transactions_icd);
@@ -37,8 +42,7 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
 
   trit_t bundle_hash_trits[NUM_TRITS_HASH];
 
-  for (curr_tx = (iota_transaction_t *)utarray_front(bundle); curr_tx != NULL;
-       curr_tx = (iota_transaction_t *)utarray_next(bundle, curr_tx)) {
+  BUNDLE_FOREACH(bundle, curr_tx) {
     absorb_essence(
         &kerl, transaction_address(curr_tx), transaction_value(curr_tx),
         transaction_obsolete_tag(curr_tx), transaction_timestamp(curr_tx),
@@ -59,23 +63,14 @@ void finalize_bundle(bundle_transactions_t *bundle) {
   byte_t normalized_hash[HASH_LENGTH_TRYTE];
   trit_t increased_tag_trits[NUM_TRITS_TAG];
   flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
-  trit_t one[NUM_TRITS_TAG] = {
-      1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
 
   Kerl kerl = {};
   init_kerl(&kerl);
 
   while (!valid_bundle) {
   update_hash:
-    // bundel hash
+    // bundle hash
     calculate_bundle_hash(bundle, bundle_hash);
-    // update bundle_hash
-    BUNDLE_FOREACH(bundle, curr_tx) {
-      transaction_set_bundle(curr_tx, bundle_hash);
-    }
     // normalize
     normalize_hash(bundle_hash, normalized_hash);
     kerl_reset(&kerl);
@@ -95,11 +90,17 @@ void finalize_bundle(bundle_transactions_t *bundle) {
       }
     }
     valid_bundle = true;
+    // update bundle_hash
+    BUNDLE_FOREACH(bundle, curr_tx) {
+      transaction_set_bundle(curr_tx, bundle_hash);
+    }
   }
 }
 
 // for debugging
 void dump_bundle(bundle_transactions_t *bundle) {
+#ifdef DEBUG
   iota_transaction_t *curr_tx = NULL;
   BUNDLE_FOREACH(bundle, curr_tx) { transaction_obj_dump(curr_tx); }
+#endif
 }
--- a/common/model/bundle.c
+++ b/common/model/bundle.c
@@ -29,29 +29,28 @@ void bundle_transactions_add(bundle_transactions_t *const bundle,
   utarray_push_back(bundle, transaction);
 }
 
-void bundle_calculate_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
+void bundle_calculate_hash(bundle_transactions_t *bundle, Kerl *const kerl,
+                           flex_trit_t *out) {
   iota_transaction_t *curr_tx = NULL;
   trit_t essence_trits[NUM_TRITS_ESSENCE];
-  Kerl kerl = {};
-  init_kerl(&kerl);
-
   trit_t bundle_hash_trits[NUM_TRITS_HASH];
+  init_kerl(kerl);
 
   BUNDLE_FOREACH(bundle, curr_tx) {
     absorb_essence(
-        &kerl, transaction_address(curr_tx), transaction_value(curr_tx),
+        kerl, transaction_address(curr_tx), transaction_value(curr_tx),
         transaction_obsolete_tag(curr_tx), transaction_timestamp(curr_tx),
         transaction_current_index(curr_tx), transaction_last_index(curr_tx),
         essence_trits);
   }
 
   // Squeeze kerl to get the bundle hash
-  kerl_squeeze(&kerl, bundle_hash_trits, NUM_TRITS_HASH);
+  kerl_squeeze(kerl, bundle_hash_trits, NUM_TRITS_HASH);
   flex_trits_from_trits(out, NUM_TRITS_HASH, bundle_hash_trits, NUM_TRITS_HASH,
                         NUM_TRITS_HASH);
 }
 
-void bundle_finalize(bundle_transactions_t *bundle) {
+void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
   iota_transaction_t *curr_tx = NULL;
   bool valid_bundle = false;
   iota_transaction_t *head_tx = NULL;
@@ -59,8 +58,7 @@ void bundle_finalize(bundle_transactions_t *bundle) {
   trit_t increased_tag_trits[NUM_TRITS_TAG];
   flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
 
-  Kerl kerl = {};
-  init_kerl(&kerl);
+  init_kerl(kerl);
 
   head_tx = (iota_transaction_t *)utarray_front(bundle);
   flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
@@ -69,10 +67,10 @@ void bundle_finalize(bundle_transactions_t *bundle) {
   while (!valid_bundle) {
   update_hash:
     // bundle hash
-    bundle_calculate_hash(bundle, bundle_hash);
+    bundle_calculate_hash(bundle, kerl, bundle_hash);
     // normalize
     normalize_hash(bundle_hash, normalized_hash);
-    kerl_reset(&kerl);
+    kerl_reset(kerl);
     // checking 'M'
     for (int i = 0; i < HASH_LENGTH_TRYTE; i++) {
       if (normalized_hash[i] == 13) {
--- a/common/model/bundle.c
+++ b/common/model/bundle.c
@@ -58,8 +58,6 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
   trit_t increased_tag_trits[NUM_TRITS_TAG];
   flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
 
-  init_kerl(kerl);
-
   head_tx = (iota_transaction_t *)utarray_front(bundle);
   flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
                       transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
@@ -70,7 +68,6 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
     bundle_calculate_hash(bundle, kerl, bundle_hash);
     // normalize
     normalize_hash(bundle_hash, normalized_hash);
-    kerl_reset(kerl);
     // checking 'M'
     for (int i = 0; i < HASH_LENGTH_TRYTE; i++) {
       if (normalized_hash[i] == 13) {
@@ -90,10 +87,9 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
   }
 }
 
-// for debugging
-void bundle_dump(bundle_transactions_t *bundle) {
 #ifdef DEBUG
+void bundle_dump(bundle_transactions_t *bundle) {
   iota_transaction_t *curr_tx = NULL;
   BUNDLE_FOREACH(bundle, curr_tx) { transaction_obj_dump(curr_tx); }
-#endif
 }
+#endif
--- a/common/model/bundle.c
+++ b/common/model/bundle.c
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
   flex_trits_from_trits(out, NUM_TRITS_HASH, bundle_hash_trits, NUM_TRITS_HASH,
                         NUM_TRITS_HASH);
 }
+
+void finalize_bundle(bundle_transactions_t *bundle) {
+  iota_transaction_t *curr_tx = NULL;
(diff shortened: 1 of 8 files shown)


common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+  byte_t normalized_hash[HASH_LENGTH_TRYTE];
+  trit_t increased_tag_trits[NUM_TRITS_TAG];
+  flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
+  trit_t one[NUM_TRITS_TAG] = {
+      1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
+  Kerl kerl = {};
```
th0br0: Let's have the user pass in a Kerl instance.

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+                              NUM_TRITS_TAG);
+        goto update_hash;
+      }
+    }
+    valid_bundle = true;
+  }
+}
+
+// for debugging
+void dump_bundle(bundle_transactions_t *bundle) {
```
th0br0: guard this with `#ifdef DEBUG`?

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
   flex_trits_from_trits(out, NUM_TRITS_HASH, bundle_hash_trits, NUM_TRITS_HASH,
                         NUM_TRITS_HASH);
 }
+
+void finalize_bundle(bundle_transactions_t *bundle) {
```
th0br0: could we call this `bundle_finalize`?

common/model/bundle.c
th0br0: this also goes for `calculate_bundle_hash` - less memory :)

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+    // normalize
+    normalize_hash(bundle_hash, normalized_hash);
+    kerl_reset(&kerl);
+    // checking 'M'
+    for (int i = 0; i < HASH_LENGTH_TRYTE; i++) {
+      if (normalized_hash[i] == 13) {
+        // Insecure bundle. Increment Tag and recompute bundle hash.
+        head_tx = (iota_transaction_t *)utarray_front(bundle);
+        flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
+                            transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
```
th0br0: use `transaction_set_obsolete_tag`

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+                            NUM_TRITS_TAG);
+        add_trits(one, increased_tag_trits, NUM_TRITS_TAG);
+        flex_trits_from_trits(transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
+                              increased_tag_trits, NUM_TRITS_TAG,
```
th0br0, line 77: use `transaction_set_obsolete_tag`

common/model/bundle.c
th0br0: also, you just need to get this tag *once* and then increment by one each time. no need to read every time.

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+    kerl_reset(&kerl);
+    // checking 'M'
+    for (int i = 0; i < HASH_LENGTH_TRYTE; i++) {
+      if (normalized_hash[i] == 13) {
+        // Insecure bundle. Increment Tag and recompute bundle hash.
+        head_tx = (iota_transaction_t *)utarray_front(bundle);
+        flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
+                            transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
+                            NUM_TRITS_TAG);
+        add_trits(one, increased_tag_trits, NUM_TRITS_TAG);
```
th0br0: could you make this a const / static outside of the function scope?

common/model/transaction.c
```diff
@@ -347,3 +347,65 @@ size_t transaction_deserialize_from_trits(iota_transaction_t *const transaction,
 void transaction_free(iota_transaction_t *const transaction) {
   free(transaction);
 }
+
+// for debugging
+void transaction_obj_dump(iota_transaction_t *tx_obj) {
```
th0br0: guard with `#ifdef DEBUG`

common/model/transaction.h
```diff
@@ -13,6 +13,8 @@ extern "C" {
 #define __COMMON_MODEL_TRANSACTION_H_
 
 #include <assert.h>
+#include <inttypes.h>
```
th0br0: not needed?

common/model/transaction.h
```diff
@@ -13,6 +13,8 @@ extern "C" {
...
+#include <stdio.h>
```
th0br0: not needed

th0br0 (changes requested):

common/model/transaction.h
oopsmonk: for debugging.

common/model/transaction.c
oopsmonk: sure.

common/model/bundle.c
oopsmonk: it's inefficient in this case.

common/model/bundle.c
oopsmonk: need to update the `obsolete_tag` for `calculate_bundle_hash`, isn't it?

common/model/bundle.c
oopsmonk: sure, or `bundle_finalization`? b/c finalize is a verb.

common/model/bundle.c
oopsmonk: yup, it's a quick approach to ease `prepare_transfers` implementation. I think except the Kerl, I am trying to reduce flex_trits conversions here, later. so an optimization version comes after the APIs. what do you think?

common/model/bundle.c
oopsmonk: btw, `finalize_bundle` is the similar naming rule with `calculate_bundle_hash`.

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
+  Kerl kerl = {};
+  init_kerl(&kerl);
+
+  while (!valid_bundle) {
+  update_hash:
+    // bundel hash
```
thibault-martinez: Typo

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+    calculate_bundle_hash(bundle, bundle_hash);
+    // update bundle_hash
+    BUNDLE_FOREACH(bundle, curr_tx) {
+      transaction_set_bundle(curr_tx, bundle_hash);
+    }
```
thibault-martinez: Set the bundle hash to txs only when it's valid ?

common/model/bundle.c
```diff
@@ -51,3 +51,55 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+        // Insecure bundle. Increment Tag and recompute bundle hash.
+        head_tx = (iota_transaction_t *)utarray_front(bundle);
+        flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
+                            transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
+                            NUM_TRITS_TAG);
+        add_trits(one, increased_tag_trits, NUM_TRITS_TAG);
+        flex_trits_from_trits(transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
+                              increased_tag_trits, NUM_TRITS_TAG,
+                              NUM_TRITS_TAG);
+        goto update_hash;
```
thibault-martinez, line 79: Just use `continue;` ?

thibault-martinez (changes requested):

common/model/bundle.c
oopsmonk: right, thanks :)

common/model/bundle.c
oopsmonk: nice try :-P

common/model/bundle.c
oopsmonk: done.

common/model/transfer.h
```diff
@@ -20,7 +20,10 @@ extern "C" {
 #include "common/trinary/trit_long.h"
 #include "utils/logger_helper.h"
 
-#define NUM_TRITS_ESSENCE 486
+// bundle essence = 486 trits
+#define NUM_TRITS_ESSENCE                                         \
+  (NUM_TRITS_ADDRESS + NUM_TRITS_VALUE + NUM_TRITS_OBSOLETE_TAG + \
+   NUM_TRITS_TIMESTAMP + NUM_TRITS_CURRENT_INDEX + NUM_TRITS_LAST_INDEX)
```
thibault-martinez: Can you put this in `transaction.h` with the other sizes ?

common/model/tests/test_bundle.c
```diff
@@ -33,10 +33,78 @@ void test_normalized_bundle(void) {
...
+  transaction_set_obsolete_tag(&tx, tag);
+  transaction_set_timestamp(&tx, 1546760811);
+  transaction_set_value(&tx, 0);
+  transaction_set_current_index(&tx, 2);
+  transaction_set_last_index(&tx, 2);
+  tx.loaded_columns_mask |= MASK_ALL_COLUMNS;
+  bundle_transactions_add(bundle, &tx);
+
+  bundle_finalize(bundle);
+  bundle_dump(bundle);
```
thibault-martinez, line 97: Remove

common/model/bundle.c
th0br0: yes. but you don't need to `flex_trits_to_trits` it each time, only once. (and then increment that one)

common/model/bundle.c
th0br0: oh you're right, this is trits->flex_trits in place.

common/model/transaction.h
th0br0: you can include these in the .c though? no need to depend on them for all includers of this header

common/model/bundle.c
th0br0: it's a minimal change, at least file an issue?

common/model/bundle.c
```diff
@@ -51,3 +55,52 @@ void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
...
+    kerl_reset(&kerl);
+    // checking 'M'
+    for (int i = 0; i < HASH_LENGTH_TRYTE; i++) {
+      if (normalized_hash[i] == 13) {
+        // Insecure bundle. Increment Tag and recompute bundle hash.
+        head_tx = (iota_transaction_t *)utarray_front(bundle);
+        flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
+                            transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
+                            NUM_TRITS_TAG);
+        add_trits(one, increased_tag_trits, NUM_TRITS_TAG);
```
thibault-martinez: I think you can do `add_assign(increased_tag_trits, NUM_TRITS_TAG, 1)`; hence no need for that `one` static array

th0br0 (changes requested):

thibault-martinez (changes requested):

common/model/tests/test_bundle.c
oopsmonk: why?

thibault-martinez: Assertions are sufficient, IMHO we shouldn't pollute stdout with logs if we are looking for information...

common/model/bundle.c
oopsmonk: cool, confused about the name of `add_assign`, could you explain more?

thibault-martinez: It modifies the trits you give it by operating a trit addition of an index on it.
For example, it's used to obtain a subseed by adding an index to a seed: https://github.com/iotaledger/entangled/blob/develop/common/sign/v1/iss.c.inc#L22

common/model/bundle.c
oopsmonk: sure, filed in #613 
just don't want to change consensus/bundle_validator atm and keep this PR in the scope of the bundle model. :-)

common/model/tests/test_bundle.c
oopsmonk: IMO, it only shows up when `DEBUG` flag is applied and additional info are useful in the test case.

common/model/bundle.c
oopsmonk: you're right, thanks.

common/model/tests/test_bundle.c
thibault-martinez: Whoops I overlooked, you're right

thibault-martinez (approved):

common/model/bundle.c
oopsmonk: done.

common/model/bundle.c
```diff
@@ -29,25 +29,71 @@ void bundle_transactions_add(bundle_transactions_t *const bundle,
...
+
+void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
+  iota_transaction_t *curr_tx = NULL;
+  bool valid_bundle = false;
+  iota_transaction_t *head_tx = NULL;
+  byte_t normalized_hash[HASH_LENGTH_TRYTE];
+  trit_t increased_tag_trits[NUM_TRITS_TAG];
+  flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
+
+  init_kerl(kerl);
```
th0br0: not needed. reset inside the loop via `bundle_calculate_hash`

common/model/bundle.c
```diff
@@ -29,25 +29,71 @@ void bundle_transactions_add(bundle_transactions_t *const bundle,
...
+
+  head_tx = (iota_transaction_t *)utarray_front(bundle);
+  flex_trits_to_trits(increased_tag_trits, NUM_TRITS_TAG,
+                      transaction_obsolete_tag(head_tx), NUM_TRITS_TAG,
+                      NUM_TRITS_TAG);
+  while (!valid_bundle) {
+  update_hash:
+    // bundle hash
```
th0br0, line 67: `kerl_reset(kerl)` is missing here

common/model/bundle.c
```diff
@@ -29,25 +29,71 @@ void bundle_transactions_add(bundle_transactions_t *const bundle,
...
-void calculate_bundle_hash(bundle_transactions_t *bundle, flex_trit_t *out) {
+void bundle_calculate_hash(bundle_transactions_t *bundle, Kerl *const kerl,
+                           flex_trit_t *out) {
   iota_transaction_t *curr_tx = NULL;
   trit_t essence_trits[NUM_TRITS_ESSENCE];
-  Kerl kerl = {};
-  init_kerl(&kerl);
-
   trit_t bundle_hash_trits[NUM_TRITS_HASH];
+  init_kerl(kerl);
```
th0br0, line 37: kerl_reset(kerl)

common/model/tests/test_bundle.c
```diff
@@ -33,10 +33,80 @@ void test_normalized_bundle(void) {
...
 
+void test_bundle_finalize(void) {
+  bundle_transactions_t *bundle = NULL;
+  flex_trit_t addr_sender[FLEX_TRIT_SIZE_243] = {};
+  flex_trit_t addr_receiver[FLEX_TRIT_SIZE_243] = {};
+  flex_trit_t tag[FLEX_TRIT_SIZE_81] = {};
+  flex_trit_t obsolete_tag[FLEX_TRIT_SIZE_81] = {};
+  iota_transaction_t tx;
+  iota_transaction_t *bundle_tx;
+  Kerl kerl = {};
```
th0br0, line 44: call `init_kerl` on this

common/model/transaction.c
```diff
@@ -347,3 +351,67 @@ size_t transaction_deserialize_from_trits(iota_transaction_t *const transaction,
...
```
th0br0: could we `#ifdef DEBUG` the whole method not just the impl/

common/model/transaction.h
```diff
@@ -399,6 +404,8 @@ static inline void transaction_set_arrival_timestamp(
  ***********************************************************************************************************/
 
 void transaction_reset(iota_transaction_t *const transaction);
+// for debugging
+void transaction_obj_dump(iota_transaction_t *tx_obj);
```
th0br0, line 408: ifdef debug the method here as well?

consensus/bundle_validator/bundle_validator.c
```diff
@@ -135,6 +135,7 @@ retcode_t iota_consensus_bundle_validator_validate(
   int64_t bundle_value = 0;
   flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
   bool valid_sig = true;
+  Kerl kerl = {};
```
th0br0, line 141: call `init_kerl` on this?

th0br0 (changes requested):

th0br0:
almost there! :)

common/model/bundle.c
oopsmonk: init_kerl == kerl_reset, do we actully need `kerl_reset`?

common/model/bundle.c
oopsmonk: at line#73, but it's redundant, let's just make sure kerl has initialized before it's called.

common/model/tests/test_bundle.c
oopsmonk: it's called inside the bundle method.

consensus/bundle_validator/bundle_validator.c
oopsmonk: kerl has initialized in the bundle method.

common/model/bundle.c
th0br0: no, my bad. thought there was a difference!

th0br0 (approved):

PR #666, opened 2019-01-08
utils/containers/lock_free: queue - small refac

tsvisabo:
- Generic include guards 
- remove unnecessary arg in init method
- export queue struct to commonly included header since it is not type dependent


th0br0:
superseded by #667

PR #667, opened 2019-01-09
containers: use concurrencykit MPMC FIFO wherever possible

thibault-martinez:
This PR uses lock-free MPMC FIFOs on the following components:
- gossip/processor
- gossip/broadcaster
- gossip/responder
- consensus/milestone_tracker

# Test Plan:
CI


utils/containers/lock_free/lf_mpmc_queue.h.tpl
```diff
@@ -0,0 +1,77 @@
...
+#define LF_MPMC_QUEUE_FOREACH(fifo, entry) CK_FIFO_MPMC_FOREACH(fifo, entry)
+#define LF_MPMC_QUEUE_FOREACH_SAFE(fifo, entry, T) \
+  CK_FIFO_MPMC_FOREACH_SAFE(fifo, entry, T)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// A generic concurrent lock-free Multi Producer Multi Consumer queue
+typedef ck_fifo_mpmc_t lf_mpmc_queue_{TYPE}_t;
```
tsvisabo: seems like we could do:
`typedef ck_fifo_mpmc_t lf_mpmc_queue_t;` 
this does not provide type safety, so we might as well get rid of the {TYPE} param?

tsvisabo (changes requested):

utils/containers/lock_free/lf_mpmc_queue.h.tpl
thibault-martinez: I do agree that it doesn't provide type safety as they would all come down to the same type `ck_fifo_mpmc_t`. But it does show intent and clarity when declaring a queue.

PR #672, opened 2019-01-10
utils/logger: lock depending on the log level

thibault-martinez:
At the cost of a redundant check (the log level is already checked by `logger_va`), this PR locks only when the log level actually allows to log.
Starting from log level `INFO`, there is almost no lock anymore.

# Test Plan:
CI

### DIFF
--- a/utils/logger_helper.c
+++ b/utils/logger_helper.c
@@ -27,52 +27,50 @@ retcode_t logger_helper_init() {
 }
 
 retcode_t logger_helper_destroy() {
+  logger_output_deregister(stdout);
   lock_handle_destroy(&lock);
 
   return RC_OK;
 }
 
-void logger_helper_enable(char const* const logger_id,
-                          logger_level_t const level, bool const enable_color) {
-  logger_id_t id;
+logger_id_t logger_helper_enable(char const* const logger_name,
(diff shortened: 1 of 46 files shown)


th0br0 (approved):

utils/logger_helper.c
```diff
@@ -61,12 +61,17 @@ void logger_helper_release(char const* const logger_id) {
...
                          ...) {
-  logger_id_t id;
   va_list argp;
+  logger_id_t id;
+
+  if (level < logger_output_level_get(stdout)) {
+    return;
+  }
+
+  id = logger_id_request(logger_id);
```
th0br0: You're requesting outside the lock now. Is that safe?

thibault-martinez: The lock is mainly here to synchronize stdout.
`logger_id_request` returns the id if it was already created and we never log before having created the logger so it should be ok.

th0br0 (changes requested):
Thibault Martinez [7:09 AM]
Let's not merge #672 yet, you made me doubt I need to think

th0br0 (approved):

th0br0 (approved):

PR #675, opened 2019-01-14
common: removed unnecessary kerl_init.

oopsmonk:
This PR removed unnecessary kerl_init & kerl_reset calls, and make sure kerl has initialized before calling it. 

# Test Plan:
CI pass

### DIFF
--- a/common/helpers/checksum.c
+++ b/common/helpers/checksum.c
@@ -20,7 +20,6 @@
 IOTA_EXPORT char* iota_checksum(const char* input, const size_t input_length,
                                 const size_t checksum_length) {
   Kerl kerl;
-  init_kerl(&kerl);
 
   if (checksum_length == 0) {
     return NULL;
@@ -50,7 +49,6 @@ IOTA_EXPORT flex_trit_t* iota_flex_checksum(const flex_trit_t* flex_trits,
                                             const size_t num_trits,
                                             const size_t checksum_length) {
   Kerl kerl;
-  init_kerl(&kerl);
 
   if (checksum_length == 0) {
     return NULL;
--- a/common/helpers/sign.c
+++ b/common/helpers/sign.c
@@ -49,8 +49,6 @@ IOTA_EXPORT char* iota_sign_address_gen(char const* const seed,
   iss_kerl_key_digest(key, key, key_length, &kerl);
   iss_kerl_address(key, key, security * HASH_LENGTH_TRIT, &kerl);
 
-  kerl_reset(&kerl);
-
   trits_to_trytes(key, (tryte_t*)address, HASH_LENGTH_TRIT);
   memset(key, 0, key_length * sizeof(trit_t));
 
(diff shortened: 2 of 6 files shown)


common/helpers/sign.c
```diff
@@ -133,8 +131,6 @@ IOTA_EXPORT flex_trit_t* iota_flex_sign_address_gen(
   iss_kerl_key_digest(key, key, key_length, &kerl);
   iss_kerl_address(key, key, security * HASH_LENGTH_TRIT, &kerl);
 
-  kerl_reset(&kerl);
```
th0br0: These resets are needed to make sure that we don't leak key data.

common/helpers/sign.c
```diff
@@ -49,8 +49,6 @@ IOTA_EXPORT char* iota_sign_address_gen(char const* const seed,
...
```
th0br0: Needed for security paranoia reasons

common/helpers/checksum.c
```diff
@@ -50,7 +49,6 @@ IOTA_EXPORT flex_trit_t* iota_flex_checksum(const flex_trit_t* flex_trits,
                                             const size_t num_trits,
                                             const size_t checksum_length) {
   Kerl kerl;
-  init_kerl(&kerl);
```
th0br0, line 53: `kerl_hash` doesn't init the instance, so this is required

common/helpers/checksum.c
```diff
@@ -20,7 +20,6 @@
 IOTA_EXPORT char* iota_checksum(const char* input, const size_t input_length,
                                 const size_t checksum_length) {
   Kerl kerl;
-  init_kerl(&kerl);
```
th0br0, line 23: kerl_hash doesn't init, so this is required

th0br0 (changes requested):

common/helpers/checksum.c
oopsmonk: `init_kerl` is moved into `kerl_hash` makes the initialization stay close to absorb/squeeze if possible.

common/helpers/checksum.c
oopsmonk: `init_kerl` is moved into `kerl_hash` makes the initialization stay close to absorb/squeeze if possible.

common/helpers/sign.c
oopsmonk: kerl is a local object here that should not have leaking problem.

common/helpers/sign.c
oopsmonk: kerl is a local object here that should not have leaking problem.

common/helpers/checksum.c
th0br0: you're right - didn't notice that change

common/helpers/sign.c
th0br0: It's about cleaning up the memory after storing secrets in it (zeroing it out)

oopsmonk: right, sorry. btw should we zero out signature too?

th0br0: no, the signature is public anyway, so we don't need to worry about that one.

th0br0 (approved):

PR #679, opened 2019-01-15
ciri/api: add http backend

th0br0:
Adds an HTTP server to cIRI for serving an IRI-compatible HTTP API


ciri/api/api.c
```diff
@@ -527,11 +529,113 @@ retcode_t iota_api_init(iota_api_t *const api, node_t *const node,
   return RC_OK;
 }
 
+typedef struct iota_api_http_sess_s {
+  bool valid_api_version;
+  bool valid_content_type;
+  char_buffer_t *request;
+} iota_api_http_sess_t;
```
oopsmonk: What is `sess` stand for?

ciri/api/api.c
```diff
@@ -527,11 +529,113 @@ retcode_t iota_api_init(iota_api_t *const api, node_t *const node,
...
+    return MHD_YES;
+  }
+
+  if (sess->request == NULL) {
+    // POST but no body, so we skip this request
+    ret = MHD_NO;
+    goto cleanup;
+  }
+
+  cJSON *json_obj = cJSON_Parse(sess->request->data);
```
oopsmonk: Local variable declarations should at the beginning of the function.

ciri/api/api.c
```diff
@@ -527,11 +529,113 @@ retcode_t iota_api_init(iota_api_t *const api, node_t *const node,
...
+  iota_api_t *api = (iota_api_t *)cls;
+  iota_api_http_sess_t *sess = *ptr;
+  struct MHD_Response *response;
+
+  if (strncmp(method, MHD_HTTP_METHOD_POST, 4) != 0) {
+    return MHD_NO;
+  }
+
+  if (sess == NULL) {
+    sess = calloc(1, sizeof(iota_api_http_sess_t));
```
tsvisabo: what if OOM?

ciri/api/api.c
```diff
@@ -527,11 +529,113 @@ retcode_t iota_api_init(iota_api_t *const api, node_t *const node,
...
+    cJSON_Delete(json_obj);
+    ret = MHD_NO;
+    goto cleanup;
+  }
+
+  response = MHD_create_response_from_buffer(strlen(json_item->valuestring),
+                                             json_item->valuestring,
+                                             MHD_RESPMEM_MUST_COPY);
+  MHD_add_response_header(response, MHD_HTTP_HEADER_CONTENT_TYPE,
+                          "application/json");
```
tsvisabo: I'm not sure i'm understanding the flow of this callback function,
Maybe what's missing here is the routing of the command to the actual command impl?
where is the response being queued to in `MHD_queue_response`? i.e, who's the client?

tsvisabo (dismissed):

ciri/api/api.c
th0br0: this method is called at least 3 times in total (connection establishing with header parsing, payload parsing (repeatedly), final processing)
it's a weird impl quirk of libmicrohttpd

cclient/serialization/json/find_transactions.c
```diff
@@ -68,25 +68,12 @@ retcode_t json_find_transactions_serialize_request(
...
-
-  json_item = cJSON_GetObjectItemCaseSensitive(json_obj, "error");
-  if (cJSON_IsString(json_item) && (json_item->valuestring != NULL)) {
-    log_error(json_logger_id, "[%s:%d] %s %s\n", __func__, __LINE__,
-              STR_CCLIENT_RES_ERROR, json_item->valuestring);
-    cJSON_Delete(json_obj);
-    return RC_CCLIENT_RES_ERROR;
-  }
+  log_debug(json_logger_id, "[%s:%d] %s\n", __func__, __LINE__, obj);
+  JSON_CHECK_ERROR(json_obj, json_item, json_logger_id);
```
thibault-martinez, line 76: `json_item` could be declared in the `do {} while (0);` scope of the `JSON_CHECK_ERROR` macro

ciri/api/api.h
```diff
@@ -38,8 +39,7 @@ typedef struct iota_api_s {
   bool running;
   node_t *node;
   iota_consensus_t *consensus;
-  serializer_t serializer;
-  serializer_type_t serializer_type;
+  tangle_t *tangle;
```
thibault-martinez: We don't need to hold a tangle reference here. All the API methods that need a database connection already have a tangle as parameter. Also, we can't really hold one since we have many tangle connection and it all depends on who will call the method.

ciri/conf.c
```diff
@@ -137,7 +137,7 @@ static retcode_t set_conf_value(iota_ciri_conf_t* const ciri_conf,
       api_conf->max_get_trytes = atoi(value);
       break;
     case 'p':  // --port
-      api_conf->port = atoi(value);
+      // api_conf->port = atoi(value);
```
thibault-martinez, line 140: Why is the port not an  option anymore ?

ciri/api/http.h
```diff
@@ -0,0 +1,83 @@
...
+retcode_t iota_api_http_destroy(iota_api_http_t *const api);
+
+/**
+ * Returns information about your node.
+ *
+ * @param api The API
+ * @param res The response
+ *
+ * @return a status code
+ */
```
thibault-martinez: Missing prototype ?

ciri/api/http.h
```diff
@@ -0,0 +1,83 @@
...
+#include "gossip/components/broadcaster.h"
+#include "utils/handles/thread.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct iota_api_http_s {
+  bool running;
+  iota_api_t *api;
```
thibault-martinez, line 30: `iota_api_t` also contains a `running` variable and a thread that are now obsolete ?

thibault-martinez (changes requested):

cclient/serialization/json/find_transactions.c
th0br0: That'd be an unneccessary declaration. We need `json_item` for getting data out of the parsed json object in some deserialize methods

ciri/api/api.h
th0br0: Ah, right - tangle_t represents the storage connection.
What's the best way to get the tangle_t instance then though for the HTTP API?

ciri/conf.c
th0br0: the port is purely HTTP specific. Adding a HTTP config object is still on the todo list to replace this.

ciri/api/http.h
th0br0: I've removed the thread variable from there - it wasn't being used anyway.

ciri/conf.c
thibault-martinez: 👍

ciri/api/http.c
```diff
@@ -31,6 +32,15 @@ static retcode_t iota_api_http_process_request(iota_api_http_t *http,
                                                const char *payload,
                                                char_buffer_t *const out) {
   retcode_t ret = RC_OK;
+
+  if (!tangle) {
+    log_error(logger_id, "Instantiating new HTTP API database connection\n");
```
thibault-martinez, line 37: `log_info`

thibault-martinez (changes requested):

thibault-martinez (approved):

thibault-martinez:
Implemented by #977

PR #681, opened 2019-01-15
mam2: replace err_t with retcode_t

thibault-martinez:
This PR replace BSU's `err_t` with entangled's `retcode_t`

# Test Plan:
CI


th0br0 (commented):
Shouldn't we move the errors we're removing over to retcode_t?

tsvisabo (approved):

PR #688, opened 2019-01-17
common: moving bundle validation to common.

oopsmonk:
This PR split bundle_validator into two parts and moving validation part to common/model/bundle.

# Test Plan:
CI pass.


common/model/bundle.c
```diff
@@ -12,6 +12,60 @@
 
 static UT_icd bundle_transactions_icd = {sizeof(iota_transaction_t), 0, 0, 0};
 
+static retcode_t validate_signature(bundle_transactions_t const *const bundle,
+                                    trit_t const *const normalized_bundle,
+                                    bool *const is_valid) {
+  iota_transaction_t *curr_tx = NULL, *curr_inp_tx = NULL;
+  Kerl address_kerl, sig_frag_kerl;
```
th0br0: we should pass the Kerl instances as an argument to this method

common/model/bundle.c
```diff
@@ -12,6 +12,60 @@
 
 static UT_icd bundle_transactions_icd = {sizeof(iota_transaction_t), 0, 0, 0};
 
+static retcode_t validate_signature(bundle_transactions_t const *const bundle,
```
th0br0: you're validating multiple signatures at once, so cal this `validate_signatures`?

common/model/bundle.c
```diff
@@ -87,6 +141,82 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
...
+                           bundle_status_t *const status) {
+  retcode_t res = RC_OK;
+  iota_transaction_t *curr_tx = NULL;
+  int64_t index = 0, last_index = 0;
+  int64_t bundle_value = 0;
+  flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
+  bool valid_sig = true;
+  Kerl kerl = {};
+
+  *status = BUNDLE_VALID;
```
th0br0: just in case of a bug, we should default to marking the bundle as invalid?

th0br0 (changes requested):
nt

common/model/bundle.c
oopsmonk: sure

common/model/bundle.c
oopsmonk: passing the Kerl from `bundle_validator` or the caller of `bundle_validator`?

common/model/bundle.c
oopsmonk: I think should remove this, the default should be inited by the caller.

common/model/bundle.c
oopsmonk: actually, I don't see a problem about Kerl instances, this's a private method.

th0br0: Having two Kerl instances here + the one in the caller method will mean 3 * 1.4kB memory vs 2 * 1.4kB if we can pass in an instance here.

common/model/bundle.c
```diff
@@ -12,6 +12,59 @@
 
 static UT_icd bundle_transactions_icd = {sizeof(iota_transaction_t), 0, 0, 0};
 
+static retcode_t validate_signatures(bundle_transactions_t const *const bundle,
```
th0br0, line 27: Could you add inline or method level documentation to this method please?

common/model/bundle.c
```diff
@@ -87,6 +140,81 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
...
+    *status = BUNDLE_INCOMPLETE;
+    return res;
+  }
+
+  memcpy(bundle_hash, transaction_bundle(curr_tx), FLEX_TRIT_SIZE_243);
+
+  BUNDLE_FOREACH(bundle, curr_tx) {
+    bundle_value += transaction_value(curr_tx);
+
+    if (llabs(bundle_value) > MAX_IOTA_SUPPLY) {
```
th0br0, line 191: we probably want to do this chack for the `transaction_value` as well, not just the bundle value which is an aggregate?

th0br0 (changes requested):

common/model/bundle.c
oopsmonk: could you explain more? what is the purpose for adding `inline` here?

common/model/bundle.c
oopsmonk: The `bundle_validator` needs 3 Kerl instances whatever pass it or not.

common/model/bundle.c
oopsmonk: I am confused why this buggy implementation was approved. turn out I need to fix it because I moved it.

common/model/bundle.c
th0br0: not `inline`, just was referring to documentation inside the method where it makes sense

common/model/bundle.c
```diff
@@ -87,6 +140,81 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
...
+    if (transaction_current_index(curr_tx) == last_index) {
+      flex_trit_t bundle_hash_calculated[FLEX_TRIT_SIZE_243];
+      trit_t normalized_bundle[HASH_LENGTH_TRIT];
+
+      if (bundle_value != 0) {
+        *status = BUNDLE_INVALID_VALUE;
+        break;
+      }
+
+      bundle_calculate_hash(bundle, &kerl, bundle_hash_calculated);
```
th0br0: the kerl that you're using here

common/model/bundle.c
```diff
@@ -87,6 +140,81 @@ void bundle_finalize(bundle_transactions_t *bundle, Kerl *const kerl) {
...
+      if (memcmp(bundle_hash, bundle_hash_calculated, FLEX_TRIT_SIZE_243) !=
+          0) {
+        *status = BUNDLE_INVALID_HASH;
+        break;
+      }
+
+      normalize_hash_trits(bundle_hash_calculated, normalized_bundle);
+
+      res = validate_signatures(bundle, normalized_bundle, &valid_sig);
```
th0br0: can be reused here as we've already calculated the hash.

th0br0 (changes requested):

th0br0 (approved):

PR #710, opened 2019-01-23
Common/transaction: add more granularity in mask of loaded fields

### LINKED ISSUE #683: common: loaded_columns_mask in transaction object.
The `loaded_columns_mask` flag is not set in transaction_set_* methods. 
E.g. 
add `MASK_DATA` in `transaction_set_message`.
clear `loaded_columns_mask` in `transaction_reset`

Maybe we should extend it to 32 bits and check all fields in the transaction object, so we can check all fields are set in a proper way.
E.g. 
```
MASK_ESSENCE = MASK_ESSENCE_ADDRESS | 
   MASK_ESSENCE_VALUE | MASK_ESSENCE_OBSELETE_TAG 
   ....
```

francescolavra:
In a partially loaded transaction, each of the fields of the iota_transaction_t structure can be partially populated, so we need to be able to describe what is actually loaded in a more granular way than allowed by the existing field_mask_e enum. For this reason, the enum is being replaced by a new set of enums, each of which indicates which sub-field in a specific field of a transaction object is populated.
Each of the transaction_set_*() functions now sets a specific bitmask in the corresponding enum, and each of the getter functions verifies via assert() that the bitmask for the specific transaction field is set.
Resolves #683.

# Test Plan:
bazel test //...


tsvisabo (approved):

thibault-martinez:
Hi @francescolavra , thanks for the PR !

PR #753, opened 2019-01-30
consensus/milestone_validator: load obsolete tag only for candidate

### LINKED ISSUE #600: Milestone validator only needs to load obsolete tag from candidates
The milestone validator currently loads the full transaction for every candidates but only the obsolete tag is actually being used, allowing us to fetch only this field.

francescolavra:
When validating a new milestone candidate, the corresponding transaction needs to be loaded from the tangle, but only the transaction obsolete tag is actually needed. Thus, we can avoid loading unnecessary fields in the transaction by loading the obsolete tag only.

Resolves #600.

# Test Plan:

Launch ciri, wait for new milestone candidate to be received and validated.

### DIFF
--- a/common/storage/sql/sqlite3/connection.c
+++ b/common/storage/sql/sqlite3/connection.c
@@ -87,6 +87,10 @@ static retcode_t prepare_statements(sqlite3_connection_t* const connection) {
   ret |= prepare_statement(connection->db,
                            &connection->statements.transaction_select_metadata,
                            iota_statement_transaction_select_metadata);
+  ret |= prepare_statement(
+      connection->db,
+      &connection->statements.transaction_select_obsolete_tag,
+      iota_statement_transaction_select_obsolete_tag);
   ret |= prepare_statement(connection->db,
                            &connection->statements.milestone_insert,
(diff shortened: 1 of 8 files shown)


thibault-martinez:
Hi @francescolavra, thanks for the PR.

Here are some preliminary thoughts on this:

- When #600 was opened, we didn't have the different fields structures so we had to fetch the entire transaction in order to get the obsolete_tag. Things have changed and we can now fetch only the essence, which overhead is I think negligible.

- This happens only on milestone candidates verification so it's not under heavy usage.

- There is a non negligible space overhead on every connection (~10).

Overall this looks good to me, @th0br0 what do you think ?

thibault-martinez:
Hi @francescolavra, a lot of things changed since and I don't think this issue is relevant anymore. I'll close the issue now and the PR in a couple of days, unless you think it should really be done and . then we can talk about it.

francescolavra:
Hi @thibault-martinez, I'm fine with this PR being closed.

thibault-martinez:
Thanks, sorry about that.

PR #754, opened 2019-01-30
mam2: extract and clean channels and endpoints from the MAM layer

thibault-martinez:
# Test Plan:
CI

### DIFF
--- a/mam/v2/mam/alloc.h
+++ b/mam/v2/mam/alloc.h
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * MAM is based on an original implementation & specification by apmi.bsu.by
+ * [ITSec Lab]
+ *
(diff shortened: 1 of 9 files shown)


mam/v2/mam/alloc.h
```diff
@@ -0,0 +1,36 @@
...
+ * MAM allocator
+ *
+ * @field create_sponge Allocator for sponge interfaces used by
+ * channels/endpoints (WOTS, PB3 sponge, PB3 fork sponge)
+ * @field destroy_sponge Deallocator
+ */
+typedef struct mam_ialloc_s {
+  sponge_t *(*create_sponge)();
+  void (*destroy_sponge)(sponge_t *);
+} mam_ialloc_t;
```
jserv, line 30: Why was the name `ialloc` kept? How about using straightforward `mam_alloc_s`?

thibault-martinez: Because it will probably disappear in a near future

tsvisabo (approved):

PR #756, opened 2019-01-31
cclient: fix include prefixes

### LINKED ISSUE #719: Fix cclient include prefixes
At the moment, we enforce a custom include_prefix for all request/response/types

This makes using cclient from different build environments (or as a library with headers) much harder than it has to be, so we should clean this up.

wusyong:
Remove include prefixes for types/request/response.

Fix #719 

# Test Plan:
CI


th0br0:
Hi @wusyong  - I'm closing this PR in favour of #757 
We ended up doing the same change in parallel apparently, so thanks for contributing nonetheless!!!

PR #768, opened 2019-02-01
mam2: state struct

thibault-martinez:
This PR provides a minimal wallet struct that will eventually contain everything a user needs to operate MAM2.

# Test Plan:
CI


th0br0:
Could we call this something other than wallet? `mam2_state` or so? wallet implies funds, which isn't the case for MAM2

tsvisabo (commented):
Hi, could we list the missing API/fields on `mam_state`?
i.e:
- NTRU keys
- PRNG/s
- channles/endpoints
- send/recv
- serialize/deserialize state

tsvisabo (approved):

PR #803, opened 2019-02-09
Optimized curl transform formulas

muXxer:
powsrv.io Team optimized the curl transform formulas.

On my machine it's around 7% faster than before.

# Test Plan:
To benchmark the results I changed the test_curlp_ptrit.c locally on my machine:
```
int main(void) {
  UNITY_BEGIN();

  RUN_TEST(test_curl_p_27_works);
  for(size_t i = 0; i < 50000; i++)
  {
    RUN_TEST(test_curl_p_81_works);
  } 

  return UNITY_END();
}
```

### DIFF
--- a/common/curl-p/ptrit.c
+++ b/common/curl-p/ptrit.c
@@ -9,17 +9,16 @@
 #include "utils/forced_inline.h"
 
 static FORCED_INLINE void ptrit_sbox(ptrit_t *const c, ptrit_t const *const s) {
-  ptrit_s alpha, beta, gamma, delta;
+  ptrit_s alpha, beta, delta;
   size_t i = 0;
 
   for (; i < STATE_LENGTH; ++i) {
     alpha = s[CURL_INDEX[i]].low;
     beta = s[CURL_INDEX[i]].high;
-    gamma = s[CURL_INDEX[i + 1]].high;
-    delta = (alpha | (~gamma)) & (s[CURL_INDEX[i + 1]].low ^ beta);
+    delta = beta ^ s[CURL_INDEX[i + 1]].low;
 
-    c[i].low = ~delta;
-    c[i].high = (alpha ^ gamma) | delta;
+    c[i].low = ~(delta & alpha);
+    c[i].high = delta | (alpha ^ s[CURL_INDEX[i + 1]].high);
   }
 }
 
--- a/common/curl-p/ptrit.c
+++ b/common/curl-p/ptrit.c
@@ -17,7 +17,7 @@ static FORCED_INLINE void ptrit_sbox(ptrit_t *const c, ptrit_t const *const s) {
     beta = s[CURL_INDEX[i]].high;
     delta = alpha & (s[CURL_INDEX[i + 1]].low ^ beta);
 
-    c[i].low = ~(delta & alpha);
+    c[i].low = ~delta;
     c[i].high = (alpha ^ s[CURL_INDEX[i + 1]].high) | delta;
   }
 }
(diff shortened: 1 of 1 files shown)


thibault-martinez:
Some benchmarks done in https://github.com/iotaledger/iri/pull/1330

common/curl-p/ptrit.c
```diff
@@ -9,17 +9,16 @@
...
     alpha = s[CURL_INDEX[i]].low;
     beta = s[CURL_INDEX[i]].high;
-    gamma = s[CURL_INDEX[i + 1]].high;
-    delta = (alpha | (~gamma)) & (s[CURL_INDEX[i + 1]].low ^ beta);
+    delta = beta ^ s[CURL_INDEX[i + 1]].low;
 
-    c[i].low = ~delta;
-    c[i].high = (alpha ^ gamma) | delta;
+    c[i].low = ~(delta & alpha);
+    c[i].high = delta | (alpha ^ s[CURL_INDEX[i + 1]].high);
```
Wollac:
```suggestion
    c[i].high = (alpha ^ s[CURL_INDEX[i + 1]].high) | delta;
```
The formula for `c[i].high` remains unchanged.

common/curl-p/ptrit.c
```diff
@@ -9,17 +9,16 @@
...
-  ptrit_s alpha, beta, gamma, delta;
+  ptrit_s alpha, beta, delta;
   size_t i = 0;
 
   for (; i < STATE_LENGTH; ++i) {
     alpha = s[CURL_INDEX[i]].low;
     beta = s[CURL_INDEX[i]].high;
-    gamma = s[CURL_INDEX[i + 1]].high;
-    delta = (alpha | (~gamma)) & (s[CURL_INDEX[i + 1]].low ^ beta);
+    delta = beta ^ s[CURL_INDEX[i + 1]].low;
```
Wollac:
```suggestion
    delta = alpha & (s[CURL_INDEX[i + 1]].low ^ beta);
```

common/curl-p/ptrit.c
```diff
@@ -9,17 +9,16 @@
...
 
-    c[i].low = ~delta;
-    c[i].high = (alpha ^ gamma) | delta;
+    c[i].low = ~(delta & alpha);
```
Wollac:
```suggestion
    c[i].low = ~delta;
```

Wollac (commented):
The changes reduce the number of bitwise operations, while still being correct under the assumption that `.low` and `.high` never *both* have a 0-bit at the same index, which is exactly a characteristic of the `ptrit` encoding.
This assumption is correct, if e.g. `trits_to_ptrits` is used for the conversion and it is a loop invariant in `ptrit_sbox`, but it is never explicitly validated before entering this function.

I proposed a few changes to the order, to make it clearer which parts of the computation have actually been changed.
These changes can be validated more formally for every bit:

review

common/curl-p/ptrit.c
tsvisabo: Hi @muXxer , thanks for this PR, can you check if the performance gain is still the same after @Wollac  order changes suggestions and integrate them to your PR?

tsvisabo (changes requested):

muXxer:
Performance gain is the same with wollacs changes.

tsvisabo (dismissed):

tsvisabo:
@muXxer can you rebase and i'll re-approve?

muXxer:
@tsvisabo done :)

tsvisabo (approved):

Issue #762, opened 2019-02-01
network: move into utils/handles


th0br0:
We currently only have a (somewhat) generic `sockets` implementation inside `/network`. 
This should be moved into `utils/handles`.

thibault-martinez:
Closed by #804

PR #804, opened 2019-02-10
utils: move sockets to handles

howjmay:
Move sockets.c and sockets.h to utils/handles and
the change the including path in the file that includes
sockets.h


howjmay:
Fixes #762

thibault-martinez (changes requested):
Hi @HowJMay , thanks for the PR !
While you're at it, could you rename the file to `utils/handles/socket.h` (singular) to match the other handles ?

thibault-martinez (approved):

PR #809, opened 2019-02-12
infra: keeping system and IDE files in the global gitignore.

### LINKED ISSUE #800: General: Remove IDE specific gitignore
It is the best practice to not include IDE specific directories into the .gitignore file of the repository. 
The gitignore should be independent in any way. Therefore, it should not include any information about the operation system or IDE.
Better: Use a local global .gitignore. Github has also an [article](https://help.github.com/articles/ignoring-files/) about this. 

What do you think about removing everything which is not necessary for the repository?

oopsmonk:
This PR fixed #800 

# Test Plan:
No need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,42 +1,9 @@
-*.class
 .#*
-
-# Mobile Tools for Java (J2ME)
-.mtj.tmp/
 
 # Package Files #
 *.jar
 *.war
 *.ear
-
-# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
-hs_err_pid*
-
-# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and Webstorm
-# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
-
-# User-specific stuff:
-.idea/**/workspace.xml
-.idea/**/tasks.xml
-
-# Sensitive or high-churn files:
-.idea/**/dataSources/
-.idea/**/dataSources.ids
-.idea/**/dataSources.xml
-.idea/**/dataSources.local.xml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,43 +1,10 @@
-*.class
 .#*
 
-# Mobile Tools for Java (J2ME)
-.mtj.tmp/
-
 # Package Files #
 *.jar
```
oopsmonk: needed?

.gitignore
```diff
@@ -46,21 +13,6 @@ hs_err_pid*
...
-# JIRA plugin
-atlassian-ide-plugin.xml
-
-# Crashlytics plugin (for Android Studio and IntelliJ)
-com_crashlytics_export_strings.xml
-crashlytics.properties
-crashlytics-build.properties
-fabric.properties
-
 buck-out/
```
oopsmonk: what are those?

th0br0: from pre-bazel, can be removed.

.gitignore
```diff
@@ -76,11 +24,8 @@ pom.xml
...
 
-# Atom conf
-.clang_complete
-
 # Database
 ciri/ciri.db*
 
-# Visual Studio Code conf
+# ignore all VisualStudioCode user settings
 .vscode/
```
thibault-martinez, line 8: What is the reason some are removed and some aren't ?

oopsmonk: The global .gitignore genarated from `www.gitignore.io` will commit user settings, but I don't think we should keep user those settings. 

```
### VisualStudioCode ###
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
```

thibault-martinez (dismissed):

thibault-martinez (approved):

PR #821, opened 2019-02-17
mam2/api: bundle read

tsvisabo:
- API


mam/v2/api/api.c
```diff
@@ -166,12 +168,60 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
 
-retcode_t mam_api_bundle_read(mam_api_t *const api,
-                              mam_channel_t const *const cha,
-                              bundle_transactions_t const *const bundle,
-                              flex_trit_t *const payload) {
-  if (api == NULL || cha == NULL || bundle == NULL || payload == NULL) {
-    return RC_NULL_PARAM;
+bool mam_api_bundle_contains_header(bundle_transactions_t const *const bundle) {
+  iota_transaction_t *tx = (iota_transaction_t *)utarray_front(bundle);
+  return flex_trits_are_null(transaction_tag(tx), NUM_TRITS_TAG);
```
thibault-martinez: Size is in flex_trits

mam/v2/api/api.c
```diff
@@ -166,12 +168,60 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+}
+
+retcode_t mam_api_bundle_read_msg(mam_recv_msg_context_t *const ctx,
+                                  mam_channel_t const *const cha,
```
thibault-martinez: Not needed

mam/v2/api/api.c
```diff
@@ -166,12 +168,60 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+                                  bundle_transactions_t const *const bundle,
+                                  flex_trit_t **const packet_payload,
+                                  trits_t session_key, trits_t msg_id) {
```
thibault-martinez: Why is session_key a parameter ?

mam/v2/api/api.c
```diff
@@ -166,12 +168,60 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
-                              bundle_transactions_t const *const bundle,
-                              flex_trit_t *const payload) {
-  if (api == NULL || cha == NULL || bundle == NULL || payload == NULL) {
-    return RC_NULL_PARAM;
+bool mam_api_bundle_contains_header(bundle_transactions_t const *const bundle) {
+  iota_transaction_t *tx = (iota_transaction_t *)utarray_front(bundle);
+  return flex_trits_are_null(transaction_tag(tx), NUM_TRITS_TAG);
+}
+
+retcode_t mam_api_bundle_read_msg(mam_recv_msg_context_t *const ctx,
```
thibault-martinez: I think we should have `mam_api_bundle_read_header` and `mam_api_bundle_read_packet`. And `mam_api_bundle_read_msg` would call them depending on if it finds a header or not. Otherwise we'll have duplicate code in `mam_api_bundle_read_packet`.

thibault-martinez (changes requested):

mam/v2/api/api.c
tsvisabo: because when you want to read the rest of the packets that belong to that msg, you need it for decryption

mam/v2/api/api.c
tsvisabo: yes

mam/v2/api/api.c
thibault-martinez: You don't actually need it. You absorb it into your spongos state and you just need to keep your spongos.

tsvisabo: Ok then :)

tsvisabo: everything in MAM is soo cryptic

thibault-martinez: :trollface:

mam/v2/mam/mam.h
```diff
@@ -0,0 +1,172 @@
+/*
```
thibault-martinez: This file has been removed/renamed to message.h

mam/v2/mam/tests/test_message.c
```diff
@@ -142,7 +142,10 @@ static void message_test_generic_receive_msg(
     trits_copy(mam_channel_id(cha), mam_msg_recv_cfg_chid(cfg));
   }
 
+  trit_t session_key_trits[MAM2_SPONGE_KEY_SIZE];
+  trits_t session_key = trits_from_rep(MAM2_SPONGE_KEY_SIZE, session_key_trits);
```
thibault-martinez: Not needed

mam/v2/examples/recv.c
```diff
@@ -0,0 +1,141 @@
...
+  if (mam_api_init(&api, SENDER_SEED) != RC_OK) {
+    fprintf(stderr, "mam_api_init failed\n");
+    return EXIT_FAILURE;
+  }
+
+  trits_t cha_name = trits_alloc(3 * strlen(TEST_CHANNEL_NAME));
+  trits_from_str(cha_name, TEST_CHANNEL_NAME);
+  mam_channel_t *cha = malloc(sizeof(mam_channel_t));
+  memset(cha, 0, sizeof(mam_channel_t));
+  mam_channel_create(&api.prng, TEST_MSS_DEPTH, cha_name, cha);
```
thibault-martinez: Receiver doesn't need to create channels

mam/v2/api/api.h
```diff
@@ -57,10 +59,51 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+retcode_t mam_api_bundle_read_msg(mam_api_t *const api,
+                                  mam_channel_t const *const cha,
+                                  bundle_transactions_t const *const bundle,
+                                  flex_trit_t **const packet_payload);
+
+/**
+ * Reads next packet
+ *
+ * @param api - The API
+ * @param cha - The channel the message belongs to
```
thibault-martinez: remove

mam/v2/api/api.c
```diff
@@ -52,7 +57,10 @@ retcode_t mam_api_destroy(mam_api_t *const api) {
   mam_ntru_sk_t_set_free(&api->ntru_sks);
   mam_ntru_pk_t_set_free(&api->ntru_pks);
   mam_psk_t_set_free(&api->psks);
-  ret = trit_t_to_mam_msg_send_context_t_map_free(&api->send_ctxs);
+  ERR_BIND_RETURN(trit_t_to_mam_msg_send_context_t_map_free(&api->send_ctxs),
+                  ret);
+  ERR_BIND_RETURN(trit_t_to_mam_msg_recv_context_t_map_free(&api->recv_ctxs),
+                  ret);
```
thibault-martinez: We can't return if one fail, it would mean that we doesn't free the others.
Also, our usage `ERR_BIND_RETURN` across MAM is very inconsistent, not sure what to do with that.

mam/v2/api/api.c
thibault-martinez: Not fixed

thibault-martinez (changes requested):

mam/v2/examples/recv.c
tsvisabo: That's a way to generate the channel's id from prng and channel name

mam/v2/api/api.c
tsvisabo: true

mam/v2/examples/recv.c
thibault-martinez: Yes, but in real life we would just send the channel id to the receiver right ?

tsvisabo: Maybe, i'll add it as a hard coded value in the test

mam/v2/api/api.c
```diff
@@ -183,12 +198,94 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+
+static retcode_t mam_api_bundle_read_header(
+    mam_msg_recv_context_t const *const ctx, trits_t *const msg) {
+  retcode_t err;
+  ERR_BIND_RETURN(mam_msg_recv(ctx, msg), err);
+  return err;
+}
+
+retcode_t mam_api_bundle_read_msg(mam_api_t *const api,
+                                  flex_trit_t const *const ch_id,
```
thibault-martinez: Do we actually need to pass the ch_id ? Since it the address of all txs in the bundle ?

mam/v2/api/api.c
```diff
@@ -183,12 +198,94 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+                                  bundle_transactions_t const *const bundle,
+                                  flex_trit_t **const packet_payload) {
+  retcode_t err = RC_OK;
+  MAM2_ASSERT(packet_payload && *packet_payload == NULL);
+  if (!mam_api_bundle_contains_header(bundle)) {
```
thibault-martinez, line 238: Why do we return if there is no header ? That won't work on single packets

mam/v2/api/api.c
```diff
@@ -183,12 +198,94 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+    *packet_payload = malloc(NUM_FLEX_TRITS_FOR_TRITS(num_trits_in_packet));
+    flex_trits_from_trits(*packet_payload, num_trits_in_packet, packet_trits,
+                          num_trits_in_packet, num_trits_in_packet);
+  }
+
+  return trit_t_to_mam_msg_recv_context_t_map_add(
+      &api->recv_ctxs, mam_msg_recv_cfg_msg_id(&cfg).p, cfg);
 }
+
+retcode_t mam_api_bundle_read_packet(mam_api_t const *const api,
```
thibault-martinez, line 285: What's the status of this function ? Unused ?

thibault-martinez (changes requested):

mam/v2/api/api.c
tsvisabo: Nope, only if we want to test that the addresses are the same, OTOH, the bundles should already be filtered by the address (channel_id)

mam/v2/api/api.c
tsvisabo: Of course it won't, single packets are going to be read with `mam_api_bundle_read_packet`

mam/v2/api/api.c
tsvisabo: not yet implemented

mam/v2/api/api.c
```diff
@@ -183,12 +198,112 @@ retcode_t mam_api_bundle_write_packet(mam_api_t *const api,
...
+                    err);
   }
+
   return RC_OK;
 }
+
+retcode_t mam_api_bundle_read_packet(mam_api_t const *const api,
+                                     bundle_transactions_t const *const bundle,
+                                     flex_trit_t **const packet_payload,
+                                     trit_t const *const msg_id,
```
thibault-martinez: msg_id is not needed since it's on the tag. Otherwise we have to ask the user to provide it.

thibault-martinez (changes requested):

thibault-martinez (approved):

Issue #835, opened 2019-02-21
mam2: various notes and bugs


thibault-martinez:
- [x] `mam_wots_t` remove `spongos` field, it is volatile and can be instantiated locally in functions `mam_wots_xxx`. Rename `mam_wots_t` to `mam_wots_sk_t`.
- [x] `mam_channel_t` `name` convert to a _fixed size_ field like `trit_t name[MAM2_CHANNEL_NAME_SIZE]`; fixed size will have an advantage that the prefix can be statically defined; same for `mam_endpoint_t`.
- [x] TODO: update the way nonces are absorbed.
- [x]  TODO: add trusted pks.
- [x]  TODO: automatically deal with `chid1`, `epid1` - make them trusted?
- [x]  `ERR_GUARD_RETURN` don't need third `e` argument - it can directly `return err;`. Thus many functions won't need `e` variable.
- [x]  Rename `enpoint` typo to `endpoint`.
- [x]  Use `prng_destination_tryte_t` type for `destination` argument in `mam_prng_genX` functions?
- [x]  Use a secure variant of `memset` in `xxx_destroy` functions to clean secret keys. `memset` is considered unsafe and may be trown out by the compiler. See https://wiki.sei.cmu.edu/confluence/display/c/MSC06-C.+Beware+of+compiler+optimizations
- [x]  If `mam_endpoint_create` fails should `mam_endpoint_destroy` be called? If not that may lead to memleak (for example, if `mss_create` fails `endpoint->name` will not be freed).
- [x]  `trits_free` can't be called twice with the same argument; it requires careful use or manual nullification of trits (`trits_free(x); x = trits_null();`). Better change to `void trits_free(trits_t *x)` or add nullification after each call to `trits_free` (see `mam_endpoint_destroy`).
- [x]  `pb3_sizeof_size_t` takes size in _trytes_ as it usually encodes size of the following `trytes` (except when it encodes `repeated` count). In `mam_endpoint_serialized_size` it is called with `mss_size` which is size in _trits_, just call `pb3_sizeof_size_t(mss_size / 3)`. Similar in other places. In `xxx_deserialize` `pb3_decode_size_t` should return size in _trytes_ hence `trits_alloc` should be called with `3 * size`.
- [x]  Maybe make `mss_serialize` take `trits_t *buffer` and return `buffer` after the end of serialized data?
[...]

PR #839, opened 2019-02-23
Use a secure variant of `memset` 

howjmay:
#835 
memest_s has been supported since C11, so I implemented memset_safe for earlier C sandard

### DIFF
--- a/cclient/http/http.c
+++ b/cclient/http/http.c
@@ -117,7 +117,7 @@ static retcode_t read_data_from_iota_service(int sockfd,
     } else if (response_context.status == IOTA_REQUEST_STATUS_DONE) {
       return RC_OK;
     }
-    memset(buffer, 0, sizeof(buffer));
+    memset_safe(buffer, sizeof(buffer), 0, sizeof(buffer));
   }
   if (num_received <= 0) {
     return RC_CCLIENT_HTTP;
--- a/cclient/http/http.c
+++ b/cclient/http/http.c
@@ -117,7 +117,7 @@ static retcode_t read_data_from_iota_service(int sockfd,
     } else if (response_context.status == IOTA_REQUEST_STATUS_DONE) {
       return RC_OK;
     }
-    memset_safe(buffer, sizeof(buffer), 0, sizeof(buffer));
+    memset(buffer, 0, sizeof(buffer));
   }
   if (num_received <= 0) {
     return RC_CCLIENT_HTTP;
--- a/common/curl-p/curl_p_s.h
+++ b/common/curl-p/curl_p_s.h
@@ -9,6 +9,7 @@
 #define __CURL_S_H__
 
 #include "common/trinary/trits.h"
+#include "urils/memset_safe.h"
 
 #define S_STATE_LENGTH 81
 
--- a/common/curl-p/curl_p_s.h
+++ b/common/curl-p/curl_p_s.h
@@ -9,7 +9,7 @@
 #define __CURL_S_H__
 
 #include "common/trinary/trits.h"
-#include "urils/memset_safe.h"
+#include "utils/memset_safe.h"
 
 #define S_STATE_LENGTH 81
 
--- a/consensus/milestone_tracker/milestone_tracker.c
+++ b/consensus/milestone_tracker/milestone_tracker.c
@@ -439,7 +439,7 @@ retcode_t iota_milestone_tracker_destroy(milestone_tracker_t* const mt) {
 
   hash243_queue_free(&mt->candidates);
   rw_lock_handle_destroy(&mt->candidates_lock);
-  memset(mt, 0, sizeof(milestone_tracker_t));
+  memset_safe(mt, sizeof(milestone_tracker_t), 0, sizeof(milestone_tracker_t));
   logger_helper_release(logger_id);
 
   return ret;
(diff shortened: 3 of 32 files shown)


utils/memset_safe.h
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2018 IOTA Stiftung
```
jserv: Bump copyright date.

utils/memset_safe.c
```diff
@@ -0,0 +1,26 @@
...
+retcode_t memset_safe(void *dest, size_t destsz, int ch, size_t count) {
+  if (dest == NULL) {
+    return RC_OOM;
+  } else if (destsz > SIZE_MAX) {
+    return RC_OOM;
+  } else if (count > destsz) {
+    return RC_OOM;
+  }
+
+  volatile unsigned char *ptr = dest;
```
th0br0, line 27: This should default to using memset_s if `__STDC_LIB_EXT1__` is defined

cclient/http/http.c
```diff
@@ -117,7 +117,7 @@ static retcode_t read_data_from_iota_service(int sockfd,
     } else if (response_context.status == IOTA_REQUEST_STATUS_DONE) {
       return RC_OK;
     }
-    memset(buffer, 0, sizeof(buffer));
+    memset_safe(buffer, sizeof(buffer), 0, sizeof(buffer));
```
th0br0: We don't need a safe memset here

consensus/milestone_tracker/milestone_tracker.c
```diff
@@ -439,7 +439,7 @@ retcode_t iota_milestone_tracker_destroy(milestone_tracker_t* const mt) {
 
   hash243_queue_free(&mt->candidates);
   rw_lock_handle_destroy(&mt->candidates_lock);
-  memset(mt, 0, sizeof(milestone_tracker_t));
+  memset_safe(mt, sizeof(milestone_tracker_t), 0, sizeof(milestone_tracker_t));
```
th0br0: No need for safe memset

utils/containers/bitset.c
```diff
@@ -17,7 +17,8 @@ void bitset_reset(bitset_t* const bitset) {
   if (bitset->size == 0) {
     return;
   }
-  memset(bitset->raw_bits, 0, bitset->size * sizeof(*(bitset->raw_bits)));
+  memset_safe(bitset->raw_bits, bitset->size * sizeof(*(bitset->raw_bits)), 0,
```
th0br0: No need for safe memset

th0br0 (changes requested):

utils/memset_safe.h
```diff
@@ -0,0 +1,36 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef __STDC_LIB_EXT1__
+
+retcode_t memset_safe(void *dest, size_t destsz, int ch, size_t count);
+
+#else
+
```
rajivshah3: Does this need to be added?
```suggestion
#define __STDC_WANT_LIB_EXT1__ 1
```

According to https://en.cppreference.com/w/c/string/byte/memset:
> As with all bounds-checked functions, memset_s is only guaranteed to be available if `__STDC_LIB_EXT1__` is defined by the implementation and if the user defines `__STDC_WANT_LIB_EXT1__` to the integer constant 1 before including string.h.

howjmay: thank you for advice

utils/memset_safe.h
```diff
@@ -0,0 +1,37 @@
...
+#define __STDC_WANT_LIB_EXT1__ 1
```
th0br0: this is in the wrong location now. 
needs to be at top of the file (i.e. around line 11)

th0br0 (changes requested):

common/curl-p/curl_p_s.h
```diff
@@ -9,6 +9,7 @@
 #define __CURL_S_H__
 
 #include "common/trinary/trits.h"
+#include "urils/memset_safe.h"
```
thibault-martinez: Typo

utils/memset_safe.c
```diff
@@ -0,0 +1,30 @@
...
+#ifndef __STDC_LIB_EXT1__
+
+retcode_t memset_safe(void *dest, size_t destsz, int ch, size_t count) {
+  if (dest == NULL) {
+    return RC_OOM;
+  } else if (destsz > SIZE_MAX) {
+    return RC_OOM;
+  } else if (count > destsz) {
+    return RC_OOM;
+  }
```
thibault-martinez, line 25: It doesn't make sense to return `RC_OOM` (Out Of Memory) here.
We shoule return errors that are similar to the real `memset_s`.
```
     [EINVAL]           The s argument was a null pointer.

     [E2BIG]            One or both of smax or n was larger than RSIZE_MAX.

     [EOVERFLOW]        n was larger than smax.
```

thibault-martinez (dismissed):

common/curl-p/curl_p_s.h
howjmay: my fault

th0br0 (dismissed):

th0br0 (approved):

PR #842, opened 2019-02-25
mobile/android: Use API level 19 to prevent a crash on some devices

rajivshah3:
Original issue: https://github.com/iotaledger/trinity-wallet/issues/1135

In optimized builds for `armeabi-v7a`, the `memcpy` function is sometimes replaced with ARM's [`__aeabi_memcpy`](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka3934.html). Because of changes in later versions of the Android NDK, crashes can sometimes occur where this function is undefined. This is fixed by using API level 19 (see https://github.com/android-ndk/ndk/issues/126#issuecomment-334672371).

# Test Plan:
No longer crashes on Motorola Moto G4 (Android 6.0.1) and LG K3 (Android 6.0.1), which were previously affected by this issue

### DIFF
--- a/mobile/android/BUILD
+++ b/mobile/android/BUILD
@@ -5,6 +5,10 @@ cc_library(
         "jni.h",
     ],
     hdrs = ["Interface.h"],
+    defines = [
+        "APP_PLATFORM=android-19",
+        "ANDROID_NATIVE_API_LEVEL=19",
+    ],
     linkopts = ["-lm"],
     deps = ["//common/helpers"],
 )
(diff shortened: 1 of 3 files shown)


mobile/android/BUILD
```diff
@@ -5,6 +5,10 @@ cc_library(
         "jni.h",
     ],
     hdrs = ["Interface.h"],
+    defines = [
```
th0br0, line 8: Defines are broadcasted to dependencies, this should be copts.

th0br0 (changes requested):

mobile/android/BUILD
th0br0: In general, these defines aren't necessary.

th0br0:
Closing this because I'm solving via revert of #808 via #843

PR #846, opened 2019-02-26
cclient/serialization/json: refactor into multiple files

th0br0:
The current JSON impl is a large file mingling logger/helper/actual impls.

Splitting this so future work on more req/res JSON encodings becomes easier.

# Test Plan:
Builds pass.


cclient/serialization/json/json_serializer.h
```diff
@@ -10,170 +10,15 @@
 
 #include <stdlib.h>
 
+#include "cclient/serialization/json/check_consistency.h"
```
oopsmonk: remove.

thibault-martinez (approved):

PR #853, opened 2019-02-27
common/curl-p: Fix missing include

rajivshah3:
Resolves the following warning when building `//common/curl-p:trit`:

```
INFO: From Compiling common/curl-p/curl_p.c:
common/curl-p/curl_p.c:65:3: warning: implicit declaration of function 'memset_safe' is invalid in C99 [-Wimplicit-function-declaration]
  memset_safe(ctx->state, sizeof(ctx->state), 0, sizeof(ctx->state));
  ^
```

# Test Plan:
CI passes and doesn't show this warning


th0br0:
gah; thanks

th0br0 (approved):

thibault-martinez (approved):

PR #860, opened 2019-02-28
mobile/android: Fix memory leaks and zero out seed even if function returns early

rajivshah3:
Fixes memory leaks and calls `memset_safe` even if the function returns early because of an error

# Test Plan:
- Tested on Android emulator


mobile/android/Interface.cpp
```diff
@@ -133,6 +134,7 @@ Java_org_iota_mobile_Interface_iota_1sign_1address_1gen_1trits(
   trit_t const* seed = (trit_t*)env->GetByteArrayElements(jseed, 0);
 
   if ((address = iota_sign_address_gen_trits(seed, index, security)) == NULL) {
+    memset_safe((void*)seed, 81, 0, 81);
```
th0br0: these are trits

th0br0 (changes requested):
General reordering & a label would be much cleaner, but this works for now I guess.

rajivshah3:
That was fast 😄

mobile/android/Interface.cpp
rajivshah3: Yikes 🤦‍♂️

rajivshah3:
@thibault-martinez if you have a chance, could you review?

mobile/android/Interface.cpp
```diff
@@ -111,6 +111,7 @@ Java_org_iota_mobile_Interface_iota_1sign_1address_1gen_1trytes(
   char const* seed = env->GetStringUTFChars(jseed, 0);
 
   if ((address = iota_sign_address_gen_trytes(seed, index, security)) == NULL) {
+    memset_safe((void*)seed, 81, 0, 81);
```
oopsmonk, line 113: you should call `ReleaseStringUTFChars` when `seed` is not used any more otherwise, it would result in a memory leak.

oopsmonk (changes requested):

oopsmonk:
Just noticed that `interface.cpp` has memory leak problems, the get method should be paired with release in JNI spec. say `GetByteArrayElements`/`ReleaseByteArrayElements`, `GetStringUTFChars`/`ReleaseStringUTFChars`.

mobile/android/Interface.cpp
```diff
@@ -55,10 +57,12 @@ JNIEXPORT jobjectArray JNICALL Java_org_iota_mobile_Interface_iota_1pow_1bundle(
   jstring outputTxtrytes = NULL;
   size_t i = 0;
 
-  flex_trits_from_trytes(flexTrunk, NUM_TRITS_TRUNK, (tryte_t*)env->GetStringUTFChars(trunk, 0), NUM_TRYTES_TRUNK,
-                         NUM_TRYTES_TRUNK);
-  flex_trits_from_trytes(flexBranch, NUM_TRITS_BRANCH, (tryte_t*)env->GetStringUTFChars(branch, 0), NUM_TRYTES_BRANCH,
-                         NUM_TRYTES_BRANCH);
+  const char* ctrunk = env->GetStringUTFChars(trunk, 0);
```
oopsmonk: const placement. const char* -> char const* for consistency

mobile/android/Interface.cpp
```diff
@@ -68,6 +72,7 @@ JNIEXPORT jobjectArray JNICALL Java_org_iota_mobile_Interface_iota_1pow_1bundle(
     const char* txTrytes = env->GetStringUTFChars(txString, 0);
```
oopsmonk: ditto

oopsmonk (changes requested):

oopsmonk (approved):

oopsmonk:
thanks :)

PR #867, opened 2019-02-28
mam2: new tag scheme

### LINKED ISSUE #864: mam2: use new tag scheme

thibault-martinez:
Fixes #864

### DIFF
--- a/mam/v2/api/api.c
+++ b/mam/v2/api/api.c
@@ -287,6 +287,8 @@ retcode_t mam_api_bundle_read_msg(mam_api_t *const api,
   flex_trits_to_trits(ctx.pk, NUM_TRITS_ADDRESS, transaction_address(curr_tx),
                       NUM_TRITS_ADDRESS, NUM_TRITS_ADDRESS);
 
+  ctx.ord = 1;
+
   // Flatten flex_trits encoded in transaction sig_or_fragment field
   // into a single long trits_t data structure
   size_t num_trits_in_bundle =
--- a/mam/v2/api/api.c
+++ b/mam/v2/api/api.c
@@ -270,14 +270,6 @@ bool mam_api_bundle_contains_header(bundle_transactions_t const *const bundle) {
   return trits_get18(trits_from_rep(18, tag + MAM2_MSG_ID_SIZE)) == 0;
 }
 
-static retcode_t mam_api_bundle_read_header(
-    mam_msg_recv_context_t const *const ctx, trits_t *const msg,
-    mam_psk_t_set_t psks, mam_ntru_sk_t_set_t ntru_sks, trits_t msg_id) {
-  retcode_t err;
-  ERR_BIND_RETURN(mam_msg_recv(ctx, msg, psks, ntru_sks, msg_id), err);
-  return err;
-}
-
 retcode_t mam_api_bundle_read_msg(mam_api_t *const api,
                                   bundle_transactions_t const *const bundle,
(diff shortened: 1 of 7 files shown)


mam/v2/mam/message.c
```diff
@@ -547,7 +545,7 @@ void mam_msg_send(mam_msg_send_context_t *ctx, mam_prng_t *prng,
   /* wrap Header */
   {
     /*  absorb tryte msgid[27]; */
-    pb3_wrap_absorb_ntrytes(&ctx->spongos, msg, msg_id);
+    pb3_absorb_external_ntrytes(&ctx->spongos, msg_id);
```
semenov-vladyslav, line 561: Technically, this works. But why not add an extra argument to `mam_msg_send` - `trits_t tag` and
```
pb3_wrap_absorb_ntrytes(&ctx->spongos, &tag, msg_id);
```
?
It will make sense as `tag` is a part or message, it's just a different location of `msg_id` and `ord`.

mam/v2/mam/message.c
```diff
@@ -644,7 +640,12 @@ void mam_msg_send_packet(mam_msg_send_context_t *ctx,
                                      checksum, ctx->mss, trits_size(payload))));
 
   /*  absorb long trint ord; */
-  pb3_wrap_absorb_longtrint(&ctx->spongos, b, ctx->ord);
+  {
```
semenov-vladyslav, line 656: I see the following alternatives here:
- `pb3_wrap_absorb_longtrint(&ctx->spongos, &tag, ctx->ord);`
- `pb3_wrap_absorb_longtrint(&ctx->spongos, &ord, ctx->ord);`
- `pb3_absorb_external_longtrint(&ctx->spongos, ctx->ord);`
This way encodings and absorptions are localized in PB3 layer. I would use the first options as it's the only place that encodes `ord` as a part of `tag`, otherwise you need to keep in sync absorption of `ctx->ord` here and encoding of `ctx->ord` into `tag` in some other place.

mam/v2/mam/message.c
```diff
@@ -810,7 +811,12 @@ retcode_t mam_msg_recv_packet(mam_msg_recv_context_t *ctx, trits_t *b,
   tryte_t checksum = -1;
 
   /*  absorb long trint ord; */
-  ERR_BIND_RETURN(pb3_unwrap_absorb_longtrint(&ctx->spongos, b, &ctx->ord), e);
+  {
```
semenov-vladyslav, line 827: `ord` is already encoded as part of tag. Or you can `pb3_encode_longtrint` instead of `trits_put18`. Or add a new function `pb3_absorb_external_ntrytes` as it can be used in send.

mam/v2/api/api.c
```diff
@@ -52,6 +52,27 @@ static void mam_api_bundle_wrap(bundle_transactions_t *const bundle,
   bundle_reset_indexes(bundle);
 }
 
+static trits_t mam_api_bundle_unwrap(bundle_transactions_t const *const bundle,
```
tsvisabo, line 58: Hell of a conflict with 899# :(

tsvisabo (dismissed):

mam/v2/api/api.c
thibault-martinez: 899 is a draft based on this. It shouldn't conflict when rebased. I hope.

tsvisabo (approved):

PR #904, opened 2019-03-03
mam/trits: use memcpy and memset

thibault-martinez:


jserv:
How about moving these functions to header `mam/v2/trits/trits.h` as `static inline` ones?

tsvisabo (approved):

Issue #893, opened 2019-03-01
cclient - split core_api and extended_api into individual compilation units


th0br0:
These large files make incremental compilations slow and increase chances of merge conflicts once multiple developers work on them.  We should split these up similar to `cclient/serialization/json`

oopsmonk:
fixed in #914

PR #914, opened 2019-03-04
Cclient - split core_api and extended_api.

### LINKED ISSUE #892: cclient - clean up unused types & typedefs
There's a bunch of currently unused types around (e.g. `input_t` in `extended_api.h`) - we should get rid of these.

oopsmonk:
This PR fixed #892 and #893

# Test Plan:
CI pass


cclient/api/core/core_init.c
```diff
@@ -0,0 +1,19 @@
+/*
```
thibault-martinez, line 1: Can we just call this `core.c` ?

cclient/api/core/core_init.h
```diff
@@ -0,0 +1,25 @@
+/*
```
thibault-martinez, line 1: Can we just call this `core.h` ?
And maybe merge it with `core_api.h` as we will end up including both anyway ?

cclient/api/extended/extended_api.h
```diff
@@ -0,0 +1,29 @@
+/*
```
thibault-martinez, line 1: Same comment as core

thibault-martinez (changes requested):

cclient/api/core/core_init.c
oopsmonk: it does the initialization job, not contains core methods.

cclient/api/core/core_init.h
oopsmonk: The user should include `core_api.h` for core API only and `extended_api.h` for both core and extended, see `ccliet/api/BUILD`. 
`core_init.h` is included in `core_api.h`, same in extended.

cclient/api/core/core_init.c
thibault-martinez: It also contains `destroy` so I  find the name misleading

cclient/api/core/core_init.h
thibault-martinez: My point was that since `core_api.h` includes `core_init.h` and we won't  use `core_init.h`  without using `core_api.h` they should be merged

thibault-martinez (approved):

PR #919, opened 2019-03-05
mam: updated CLI examples

thibault-martinez:


mam/examples/common.h
```diff
@@ -22,6 +22,7 @@
   "9999999"
 #define TEST_CHANNEL_NAME "CHANAME"
 #define TEST_MSS_DEPTH 1
+#define MAM_FILE "mam.bin"
```
tsvisabo: can't this be provided by user at cli?

tsvisabo (changes requested):

mam/examples/common.h
thibault-martinez: It could, but I'm not sure it's important/relevant to the examples.
I can add if you think it is though.

tsvisabo (approved):

PR #931, opened 2019-03-07
follow IRI ISSinPlace.digest() implementation

ed2k:


common/sign/v1/iss.c.inc
```diff
@@ -74,9 +74,9 @@ int _ISS_PREFIX(key_digest)(trit_t *key, trit_t *digest,
 
 int _ISS_PREFIX(address)(trit_t const *const digest, trit_t *const address,
                          size_t const digest_length, HASH_STATE *const state) {
+  _HASH_PREFIX(reset)(state);
```
th0br0: Please undo; we expect the state to have been reset before calling this method.
Also, all ISS methods in entangled clean up after themselves, hence the reset at the end of the method.

common/sign/v1/iss.c.inc
```diff
@@ -106,13 +106,14 @@ int _ISS_PREFIX(sig_digest)(trit_t *const dig, trit_t const *const hash,
                             HASH_STATE *const state) {
   size_t i, j;
   trit_t *sig_start = sig, *sig_end = &sig[sig_len];
-  for (i = 0; sig < sig_end; i++, sig = &sig[HASH_LENGTH_TRIT]) {
-    for (j = 0; j < (HASH_TRYTE_VAL(hash, i) - TRYTE_VALUE_MIN); j++) {
+  for (i = 0, sig = sig_start; sig < sig_end; i++, sig = &sig[HASH_LENGTH_TRIT]) {
+    for (j = (hash[i] - TRYTE_VALUE_MIN); j-- > 0;) {
```
th0br0, line 110: No, `hash` is a trit array and not a tryte array, hence the need for the `HASH_TRYTE_VAL`m acro.

common/sign/v1/iss.c.inc
```diff
@@ -106,13 +106,14 @@ int _ISS_PREFIX(sig_digest)(trit_t *const dig, trit_t const *const hash,
...
+      _HASH_PREFIX(reset)(state);
```
th0br0: See comment regarding reset / clean state above

common/sign/v1/iss.c.inc
```diff
@@ -106,13 +106,14 @@ int _ISS_PREFIX(sig_digest)(trit_t *const dig, trit_t const *const hash,
...
       _HASH_PREFIX(absorb)(state, sig, HASH_LENGTH_TRIT);
       _HASH_PREFIX(squeeze)(state, sig, HASH_LENGTH_TRIT);
-      _HASH_PREFIX(reset)(state);
     }
   }
+  _HASH_PREFIX(reset)(state);
```
th0br0: This is not necessary if reset is the last command inside the loop

common/sign/v1/tests/test_iss_kerl.c
```diff
@@ -77,6 +79,43 @@ void test_key(trit_t *s, trit_t *k, size_t l, Kerl *c) {
...
+  Kerl* c = &kerl;
+  init_kerl(&kerl);
+
+  tryte_t tryte_seed[] = TEST_SEED2;
+  trytes_to_trits(tryte_seed, seed, 81);
+  iss_kerl_subseed(seed, subseed, 0, c);
+  iss_kerl_key(subseed, k, l, c);
+
+  tryte_t addy_trytes[l];
+  trit_t s[6561];
```
th0br0: please use the relevant #defines and don't use magic numbers in this code.

common/sign/v1/tests/test_iss_kerl.c
```diff
@@ -77,6 +79,43 @@ void test_key(trit_t *s, trit_t *k, size_t l, Kerl *c) {
...
+#define EXP_SIG                                                                \
+  "DQMAVYJTIIQUCDHRIZAPYKFCFBTHNDAMHLYEEIEBSR9REQBBVUQOSDWFSGPTZFNQQTVMQPVUDH" \
+  "DTEZVSW"
+
+void test_sig_digest() {
+  const size_t key_length = ISS_KEY_LENGTH;
+  trit_t seed[HASH_LENGTH_TRIT];
+  trit_t subseed[HASH_LENGTH_TRIT];
+  trit_t key[key_length];
+  trit_t *k = key;
```
th0br0: don't use single-letter variable names please

th0br0 (changes requested):
Hi @ed2k  - thanks for the PR.
Unfortunately, other than the test, this does not look like it really improves the code.

The test is a nice addition though, if you clean it up I'd love to merge it.

common/sign/v1/tests/test_iss_kerl.c
```diff
@@ -77,6 +79,43 @@ void test_key(trit_t *s, trit_t *k, size_t l, Kerl *c) {
   "CFOYUCXLHLSUBAEYOTAWUNRPJFA9TSJNLBFLMZQASPTVCMTFBOQQRGGQ9MRZCJWYGBORJZQWVS" \
   "BLVKBVW"
 
+#define EXP_SIG                                                                \
```
thibault-martinez: Why is it called `EXP_SIG` since it's compared to an address ?

common/sign/v1/tests/test_iss_kerl.c
```diff
@@ -77,6 +79,43 @@ void test_key(trit_t *s, trit_t *k, size_t l, Kerl *c) {
...
+
+  tryte_t tryte_seed[] = TEST_SEED2;
+  trytes_to_trits(tryte_seed, seed, 81);
+  iss_kerl_subseed(seed, subseed, 0, c);
+  iss_kerl_key(subseed, k, l, c);
+
+  tryte_t addy_trytes[l];
+  trit_t s[6561];
+  trit_t d[6561];
+  memcpy(s, k, sizeof(trit_t)*6561);
```
thibault-martinez: Apply `clang_format`

common/sign/v1/tests/test_iss_kerl.c
```diff
@@ -77,6 +79,43 @@ void test_key(trit_t *s, trit_t *k, size_t l, Kerl *c) {
...
+
+  iss_kerl_sig_digest(d, s, k, l, c);
```
thibault-martinez: You're using the start of the key as hash  for the digest ?

thibault-martinez (changes requested):

common/sign/v1/iss.c.inc
ed2k: this is main change I did. I looked up java IRI ISSinPlace.digest() implementation
java code is doing in this way.

common/sign/v1/tests/test_iss_kerl.c
ed2k: how do I apply clang_format, is it some kind of pre-commit check or some command.

common/sign/v1/tests/test_iss_kerl.c
ed2k: yes

common/sign/v1/tests/test_iss_kerl.c
thibault-martinez: `./tools/formatter`

common/sign/v1/iss.c.inc
thibault-martinez: IRI is doing it this way just because it takes a trytes hash as parameter.
Entangled takes a trits hash, that's why we have the macro.
Still, both do exactly the same thing in a different way.

common/sign/v1/tests/test_iss_kerl.c
```diff
@@ -77,6 +79,41 @@ void test_key(trit_t *s, trit_t *k, size_t l, Kerl *c) {
   "CFOYUCXLHLSUBAEYOTAWUNRPJFA9TSJNLBFLMZQASPTVCMTFBOQQRGGQ9MRZCJWYGBORJZQWVS" \
   "BLVKBVW"
 
+#define EXP_DIGEST                                                             \
+  "DQMAVYJTIIQUCDHRIZAPYKFCFBTHNDAMHLYEEIEBSR9REQBBVUQOSDWFSGPTZFNQQTVMQPVUDH" \
+  "DTEZVSW"
```
thibault-martinez, line 84: May I ask where you found this example ?

ed2k:
i implemented the same unit test in iri, will have a different patch later
to prove my point.

common/sign/v1/iss.c.inc
ed2k: I think it is using trits instead. java test is here,
https://github.com/iotaledger/iri/pull/1368

thibault-martinez: It's not. It doesn't make sense to do `hash[i] - TRYTE_VALUE_MIN` if `hash[i]` is a trit. It would mean doing {-1, 0, 1} - (-13),  which is not what we want.

ed2k:
that is why i raised this issue, the change is to match what IRI has
implemented.

thibault-martinez:
Closing this PR because it originated from another issue (#960) that is now fixed.
Also, ISS tests have been added since.
Thanks anyway !

PR #1001, opened 2019-03-21
api/http: implement endpoints

thibault-martinez:
This PR implements most HTTP API endpoints.
Some are still dependent on serializations/deserializations but placeholders have been created.


oopsmonk:
Could we rename this `http.c` file to a meaningful one? for instance, `request_parser.c` or `request_handler.c`

oopsmonk (approved):

thibault-martinez:
It's named like that because we expect a binary API too so `request_parser.c` wouldn't be specific enough.

PR #1007, opened 2019-03-22
api: add error response

### LINKED ISSUE #1005: ciri: add an error response

thibault-martinez:
Fixes #1005 

# Test Plan
Improved existing tests

### DIFF
--- a/cclient/serialization/json/error.c
+++ b/cclient/serialization/json/error.c
@@ -25,11 +25,7 @@ retcode_t json_error_serialize_response(serializer_t const *const s, error_res_t
 
   json_text = cJSON_PrintUnformatted(json_root);
   if (json_text) {
-    len = strlen(json_text);
-    ret = char_buffer_allocate(out, len);
-    if (ret == RC_OK) {
-      strncpy(out->data, json_text, len);
-    }
+    char_buffer_set(out, json_text);
     cJSON_free((void *)json_text);
   }
 
--- a/cclient/serialization/json/error.c
+++ b/cclient/serialization/json/error.c
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2019 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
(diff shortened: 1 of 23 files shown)


cclient/serialization/json/error.c
```diff
@@ -0,0 +1,38 @@
...
+    log_critical(json_logger_id, "[%s:%d] %s\n", __func__, __LINE__, STR_CCLIENT_JSON_CREATE);
+    return RC_CCLIENT_JSON_CREATE;
+  }
+
+  cJSON_AddStringToObject(json_root, "error", obj->error->data);
+
+  json_text = cJSON_PrintUnformatted(json_root);
+  if (json_text) {
+    len = strlen(json_text);
+    ret = char_buffer_allocate(out, len);
```
oopsmonk: char_buffer_set

ciri/api/api.c
```diff
@@ -78,8 +78,9 @@ static iota_api_command_t get_command(char const *const command) {
  * Public functions
  */
 
-retcode_t iota_api_get_node_info(iota_api_t const *const api, get_node_info_res_t *const res) {
-  if (api == NULL || res == NULL) {
+retcode_t iota_api_get_node_info(iota_api_t const *const api, get_node_info_res_t *const res,
+                                 error_res_t **const error) {
```
oopsmonk, line 82: why double pointer?

oopsmonk (changes requested):

ciri/api/api.c
thibault-martinez: Because the response will be allocated only when an error occurs

oopsmonk: if so, you should remove the const qualifier. And maybe make error_res_t as a private inside the API methods, so we don't need to pass it. We check retcode_t if there are errors.

thibault-martinez: The `const` qualifier is on the second pointer so it's ok since it won't be changed.
This is on the API level, not on the HTTP API level. So the error needs to be passed as parameter because it will be serialized on the HTTP API level.

oopsmonk (approved):

Issue #583, opened 2018-11-30
Validate configuration variables during parsing (file & CLI)


thibault-martinez:
All 4 configuration objects should always hold valid values

PR #1029, opened 2019-03-27
gossip: CLI option to enable/disable tip solidification

### LINKED ISSUE #1028: gossip: CLI option to enable/disable tip solidification

thibault-martinez:
Fixes #1028

### DIFF
--- a/ciri/conf.c
+++ b/ciri/conf.c
@@ -68,6 +68,17 @@ static sponge_type_t get_sponge_type(char const* const sponge_type) {
   return map[i].type;
 }
 
+static retcode_t get_true_false(char const* const input, bool* const output) {
+  if (strcmp(input, "true") == 0) {
+    *output = true;
+  } else if (strcmp(input, "false") == 0) {
+    *output = false;
+  } else {
+    return RC_CIRI_CONF_INVALID_ARGUMENTS;
+  }
+  return RC_OK;
+}
+
 static int get_conf_key(char const* const key) {
   int i = 0;
   while (cli_arguments_g[i].name != NULL && strcmp(cli_arguments_g[i].name, key) != 0) {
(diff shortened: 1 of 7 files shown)


ciri/conf.c
```diff
@@ -68,6 +68,17 @@ static sponge_type_t get_sponge_type(char const* const sponge_type) {
   return map[i].type;
 }
 
+static retcode_t get_true_false(char const* const input, bool* const output) {
+  if (strcmp(input, "true") == 0) {
+    *output = true;
+  } else if (strcmp(input, "false") == 0) {
+    *output = false;
+  } else {
+    return RC_CIRI_CONF_INVALID_ARGUMENTS;
```
oopsmonk, line 77: should we remain it to the default value or do we have a mechanism to locate which argument went wrong?

ciri/conf.c
```diff
@@ -204,13 +218,7 @@ static retcode_t set_conf_value(iota_ciri_conf_t* const ciri_conf, iota_consensu
...
       break;
     case CONF_SNAPSHOT_SIGNATURE_SKIP_VALIDATION:  // --snapshot-signature-skip-validation
-      if (strcmp(value, "true") == 0) {
-        consensus_conf->snapshot_signature_skip_validation = true;
-      } else if (strcmp(value, "false") == 0) {
-        consensus_conf->snapshot_signature_skip_validation = false;
-      } else {
-        return RC_CIRI_CONF_INVALID_ARGUMENTS;
-      }
+      ret = get_true_false(value, &consensus_conf->snapshot_signature_skip_validation);
```
oopsmonk, line 221: the `ret` will be overwritten in the other case. say case CONF_TIPS_SOLIDIFIER_ENABLED set `ret` to RC_CIRI_CONF_INVALID_ARGUMENTS, but it could be set to RC_OK in this case.

oopsmonk (changes requested):

ciri/conf.c
thibault-martinez: Parameters are processed one by one by `set_conf_value` so there is no risk of overriding `ret`.

ciri/conf.c
thibault-martinez: It will be done in #583

oopsmonk (dismissed):

oopsmonk (approved):

PR #1110, opened 2019-04-16
model: reduce the size of transaction field masks

thibault-martinez:


oopsmonk:
Would you mind adding test cases for mask fields?

thibault-martinez:
@oopsmonk That's already done here https://github.com/iotaledger/entangled/blob/develop/common/model/tests/test_transaction.c#L48

oopsmonk (approved):

PR #1122, opened 2019-04-23
utils: fix current timestamp ms on windows

### LINKED ISSUE #1116: utils: investigate current_timestamp_ms failure on Windows

thibault-martinez:
Fixes #1116


utils/time.c
```diff
@@ -21,11 +21,12 @@
 
 uint64_t current_timestamp_ms() {
 #ifdef _WIN32
-  SYSTEMTIME time;
+  __int64 wintime;
 
-  GetSystemTime(&time);
-  LONG ms = (time.wSecond * 1000) + time.wMilliseconds;
-  return ms;
+  GetSystemTimeAsFileTime((FILETIME*)&wintime);
```
oopsmonk, line 26: Do not cast a pointer to a FILETIME structure to either a ULARGE_INTEGER* or __int64* value because it can cause alignment faults on 64-bit Windows.
ref: https://docs.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-filetime

oopsmonk (changes requested):

utils/time.c
thibault-martinez: I'm doing the opposite here.
Even though they don't specify anything about that in the documentation and it appears to be ok (https://stackoverflow.com/questions/5404277/porting-clock-gettime-to-windows#comment64618848_31335254), I'll try to remove ambiguity.

tsvisabo (approved):

oopsmonk (approved):

PR #1175, opened 2019-05-10
gossip: keep ping pong alive on missing requests

thibault-martinez:
Still investigating this.


thibault-martinez:
Will be part of #1318

PR #1182, opened 2019-05-13
common: transfer refactoring.

### LINKED ISSUE #910: common - transfer model refactoring.
The transfer model is no easy to use for users, missing part:
*  validation and calculation method, ex. transfers_to_bundle
* operations for a transfer list.

oopsmonk:
Fixes #910


common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
-  flex_trit_t obsolete_tag[FLEX_TRIT_SIZE_81];
+/**
+ * @brief A transfer object
+ *
+ */
+typedef struct transfer_s {
   int64_t value;
   uint64_t timestamp;
-  void *meta;  // extend informations depended on transfer type.
+  tryte_t *message;
```
thibault-martinez, line 25: In trytes ?

common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
+}
 
-void absorb_essence(Kerl *const kerl, flex_trit_t const *address, int64_t value, flex_trit_t const *obsolete_tag,
-                    uint64_t timestamp, int64_t current_index, int64_t last_index, trit_t *const essence_trits);
-/***********************************************************************************************************
- * Transfer Iterator data structure
- ***********************************************************************************************************/
-typedef flex_trit_t *(*iota_signature_generator)(const flex_trit_t *seed, const size_t index, const size_t security,
-                                                 const flex_trit_t *bundleHash);
+static inline void signature_fragments_add(signature_fragments_t *fragments, tryte_t const *msg_or_sign) {
```
thibault-martinez, line 38: Trytes obtained with conversion, either from trits or flex_trits are not null-terminated. Is it ok to use it with `ut_str_icd` ?

thibault-martinez: More generally, we should not rely on null-termination of trytes, this is most of the time not the case.

common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
-  uint32_t current_transaction_index;
-  flex_trit_t bundle_hash[FLEX_TRIT_SIZE_243];
-  flex_trit_t *transaction_signature;
-  iota_transaction_t *transaction;
-  uint8_t dynamic_transaction;
-  iota_signature_generator iota_signature_gen;
-  transfer_t **transfers;
-} transfer_iterator_t;
+static inline void signature_fragments_free(signature_fragments_t *fragments) { utarray_free(fragments); }
+static inline tryte_t **signature_fragments_at(signature_fragments_t *fragments, int index) {
```
thibault-martinez:
```suggestion
static inline tryte_t **signature_fragments_at(signature_fragments_t *fragments, size_t index) {
```

common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
-// Free an existing transfer iterator
-void transfer_iterator_free(transfer_iterator_t **iter);
+retcode_t transfer_message_set(transfer_t *tf, tryte_t const *const msg);
+static inline void transfer_message_free(transfer_t *tf) { free(tf->message); }
+static inline tryte_t *transfer_message_get(transfer_t *tf) { return tf->message; }
 
-bool validate_output(transfer_value_out_t const *const output);
+transfer_array_t *transfer_array_new();
+static inline void transfer_array_free(transfer_array_t *transfers) { utarray_free(transfers); }
+static inline size_t transfer_array_count(transfer_array_t *transfers) { return utarray_len(transfers); }
```
thibault-martinez: const

common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
-// if `transaction` is NULL will be dynamically allocated a transaction object.
-transfer_iterator_t *transfer_iterator_new(transfer_t *transfers[], uint32_t len, Kerl *kerl,
-                                           iota_transaction_t *transaction);
+typedef UT_array transfer_array_t;
 
-// Free an existing transfer iterator
-void transfer_iterator_free(transfer_iterator_t **iter);
+retcode_t transfer_message_set(transfer_t *tf, tryte_t const *const msg);
+static inline void transfer_message_free(transfer_t *tf) { free(tf->message); }
+static inline tryte_t *transfer_message_get(transfer_t *tf) { return tf->message; }
```
thibault-martinez: const

common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
 
-bool validate_output(transfer_value_out_t const *const output);
+transfer_array_t *transfer_array_new();
+static inline void transfer_array_free(transfer_array_t *transfers) { utarray_free(transfers); }
+static inline size_t transfer_array_count(transfer_array_t *transfers) { return utarray_len(transfers); }
+static inline void transfer_array_add(transfer_array_t *tfs, transfer_t *tf) { utarray_push_back(tfs, tf); }
+static inline transfer_t *transfer_array_at(transfer_array_t *tfs, int index) {
```
thibault-martinez:
```suggestion
static inline transfer_t *transfer_array_at(transfer_array_t *tfs, size_t index) {
```

common/model/transfer.h
```diff
@@ -12,131 +12,60 @@ extern "C" {
...
+  return (transfer_t *)utarray_eltptr(tfs, index);
+}
 
-void transfer_logger_init();
-void transfer_logger_destroy();
+#define TRANSFER_NEW_FOREACH(tfs, tf) \
```
thibault-martinez: Why `NEW` ?

common/model/bundle.c
```diff
@@ -236,6 +257,76 @@ void bundle_reset_indexes(bundle_transactions_t *const bundle) {
...
 
+void bundle_set_messages(bundle_transactions_t *bundle, signature_fragments_t *messages) {
+  iota_transaction_t *tx = NULL;
+  size_t index = 0;
+  tryte_t **msg = NULL;
+  size_t msg_len = 0;
+  tryte_t trytes_buff[NUM_TRYTES_SIGNATURE];
+
+  BUNDLE_FOREACH(bundle, tx) {
+    memset(trytes_buff, '9', NUM_TRYTES_SIGNATURE);
```
thibault-martinez: We can probably use `msg_len` to avoid setting the whole array everythime

common/model/bundle.c
```diff
@@ -236,6 +257,76 @@ void bundle_reset_indexes(bundle_transactions_t *const bundle) {
...
+                           NUM_TRYTES_SIGNATURE);
+
+    memset(tx->attachment.branch, FLEX_TRIT_NULL_VALUE, FLEX_TRIT_SIZE_243);
+    memset(tx->attachment.trunk, FLEX_TRIT_NULL_VALUE, FLEX_TRIT_SIZE_243);
+    memset(tx->attachment.nonce, FLEX_TRIT_NULL_VALUE, FLEX_TRIT_SIZE_81);
+
+    transaction_set_attachment_timestamp(tx, 0);
+    transaction_set_attachment_timestamp_lower(tx, 0);
+    transaction_set_attachment_timestamp_upper(tx, 0);
+    tx->loaded_columns_mask.attachment = MASK_ATTACHMENT_ALL;
```
thibault-martinez: The name of the function`bundle_set_messages` is misleading as it suggests that we are only seting messages. Why do we need to reset all attachment fields as well ?

cclient/api/extended/prepare_transfers.c
```diff
@@ -5,27 +5,219 @@
...
+  int64_t total_transfer_value = total;
+  input_t* input = NULL;
 
-  if (transfer_iterator) {
-    for (tx = transfer_iterator_next(transfer_iterator); tx; tx = transfer_iterator_next(transfer_iterator)) {
-      bundle_transactions_add(out_bundle, tx);
+  INPUTS_FOREACH(inputs->input_array, input) {
+    int64_t input_value = input->balance;
+    int64_t substract = 0 - input_value;
+    int64_t timestamp = current_timestamp_ms();
```
thibault-martinez:
```suggestion
    uint64_t timestamp = current_timestamp_ms();
```

cclient/api/extended/prepare_transfers.c
```diff
@@ -5,27 +5,219 @@
...
+      if (remainder_value > 0 && remainder_address != NULL) {
+        // put remainder value to remainder address
+        add_transaction_entry(bundle, 1, remainder_address, remainder_value, tag, timestamp);
+        // sign and return
+        bundle_sign(bundle, seed, inputs, kerl);
+        return RC_OK;
+      } else if (remainder_value > 0) {
+        // remainder address is needed
+        address_opt_t addr_info = {0, 0, security};
+        hash243_queue_t new_addresses = NULL;
```
thibault-martinez, line 66: Free ?

thibault-martinez (changes requested):

common/model/transfer.h
oopsmonk: yes, considering of UX

common/model/transfer.h
oopsmonk: trits and flex_trits are not null-terminated, but trytes does in this case.

common/model/bundle.c
```diff
@@ -7,12 +7,33 @@
...
-#include "common/model/transfer.h"
+#include "common/helpers/sign.h"
 #include "common/trinary/trit_long.h"
 #include "common/trinary/tryte_long.h"
 
 static UT_icd bundle_transactions_icd = {sizeof(iota_transaction_t), 0, 0, 0};
 
+static void absorb_essence(Kerl *const kerl, flex_trit_t const *address, int64_t value, flex_trit_t const *obsolete_tag,
+                           uint64_t timestamp, int64_t current_index, int64_t last_index, trit_t *const essence_trits) {
+  memset(essence_trits, 0, NUM_TRITS_ESSENCE);
```
thibault-martinez, line 18: This shouldn't be necessary since we are going to rewrite on the whole array ?

common/model/bundle.c
```diff
@@ -236,6 +257,73 @@ void bundle_reset_indexes(bundle_transactions_t *const bundle) {
...
+    if (msg == NULL) {
+      break;
+    }
+    msg_len = strlen((char *)*msg);
+    memcpy(trytes_buff, *msg, msg_len);
+    memset(trytes_buff + msg_len, '9', NUM_TRYTES_SIGNATURE - msg_len);
+
+    // trytes to flex_trits
+    flex_trits_from_trytes(tx->data.signature_or_message, NUM_TRITS_SIGNATURE, trytes_buff, NUM_TRYTES_SIGNATURE,
+                           NUM_TRYTES_SIGNATURE);
```
thibault-martinez, line 278: If you `memset` `tx->data.signature_or_message` to `FLEX_TRIT_NULL_VALUE` before and then directly use `msg ` in `flex_trits_from_trytes`, you should be able to get rid of `trytes_buff` and its `memcpy`/`memset`

common/model/bundle.c
```diff
@@ -236,6 +257,73 @@ void bundle_reset_indexes(bundle_transactions_t *const bundle) {
...
+          }
+          signed_signature =
+              iota_sign_signature_gen_flex_trits(seed, input->key_index, input->security, transaction_bundle(tx));
+          if (signed_signature) {
+            // for each security level add signature
+            for (int i = 0; i < input->security; i++) {
+              tx = bundle_at(bundle, curr_index + i);
+              memcpy(tx->data.signature_or_message, signed_signature + (i * NUM_FLEX_TRITS_SIGNATURE),
+                     NUM_FLEX_TRITS_MESSAGE);
+              tx->loaded_columns_mask.data |= MASK_DATA_SIG_OR_MSG;
```
thibault-martinez, line 309: Do it only once after the loop ?

cclient/api/extended/prepare_transfers.c
```diff
@@ -5,27 +5,219 @@
...
-  retcode_t ret_code = RC_ERROR;
-  iota_transaction_t TX = {};
-  iota_transaction_t* tx = NULL;
-  transfer_iterator_t* transfer_iterator = NULL;
-  Kerl kerl = {};
+static void add_transaction_entry(bundle_transactions_t* bundle, int num_txs, flex_trit_t const* const address,
+                                  int64_t value, flex_trit_t* tag, uint64_t timestamp) {
+  iota_transaction_t tx = {};
+  for (int i = 0; i < num_txs; i++) {
+    transaction_reset(&tx);
```
thibault-martinez: Only need to reset once before the loop since all the other fields will be changed every time

cclient/api/extended/prepare_transfers.c
```diff
@@ -5,27 +5,219 @@
...
+        add_transaction_entry(bundle, 1, remainder_address, remainder_value, tag, timestamp);
+        // sign and return
+        return bundle_sign(bundle, seed, inputs, kerl);
+      } else if (remainder_value > 0) {
+        // remainder address is needed
+        address_opt_t addr_info = {0, 0, security};
+        hash243_queue_t new_addresses = NULL;
+        if ((ret = iota_client_get_new_address(serv, seed, addr_info, &new_addresses)) == RC_OK) {
+          // unused address
+          add_transaction_entry(bundle, 1, new_addresses->prev->hash, remainder_value, tag, timestamp);
```
thibault-martinez, line 69: Just confirming it's done on purpose: why `->prev->` ?

cclient/api/extended/prepare_transfers.c
```diff
@@ -5,27 +5,219 @@
...
+  }
+
+  if ((ret = iota_client_get_balances(serv, balances_req, balances_res)) == RC_OK) {
+    size_t index = 0;
+    INPUTS_FOREACH(inputs->input_array, input_elm) {
+      if (input_elm->balance != get_balances_res_balances_at(balances_res, index)) {
+        *passed = false;
+        log_error(client_extended_logger_id, "balance is not match\n");
+        goto end;
+      }
```
thibault-martinez, line 117: Not incrementing index ?

cclient/api/extended/prepare_transfers.c
```diff
@@ -5,27 +5,219 @@
...
+      if (ret == RC_OK) {
+        // add remainder
+        ret = add_remainder(serv, seed, security, &local_inputs, out_bundle, tag, total_value, remainder, &kerl);
+      }
+      inputs_clear(&local_inputs);
+    }
+  } else {
+    bundle_reset_indexes(out_bundle);
+    bundle_finalize(out_bundle, &kerl);
+    bundle_set_messages(out_bundle, sign_fragments);
```
thibault-martinez: Has already been done ?

thibault-martinez: L183

thibault-martinez (changes requested):

common/model/bundle.c
oopsmonk: it's here for ages. initialization is a good habit, IMO.

common/model/bundle.c
oopsmonk: didn't get your point, here `trytes_buff` is for padding and conversion.

common/model/bundle.c
oopsmonk: for what? it sets the flag to tx after filling.

cclient/api/extended/prepare_transfers.c
oopsmonk: it's the location of the unused address.

thibault-martinez: Thanks

common/model/bundle.c
thibault-martinez: But why to set it as many times as the security when you can set it once after ?

common/model/bundle.c
thibault-martinez: I meant that you can achieve exactly the same without the `trytes_buff `

common/model/bundle.c
thibault-martinez: Yes, it is a good habit when something can go wrong, here nothing can go wrong and the array will always be overwritten. You can resolve the conversation if you feel you want to keep it.

common/model/bundle.c
oopsmonk: No, `msg` is trytes with a random length, unable to do the in-place conversion. that's why need a `trytes_buff` which length is NUM_TRYTES_SIGNATURE for padding and trytes->flex trits conversion..

common/model/bundle.c
oopsmonk: no thing about security level, treat it as `transaction_set_message`

common/model/bundle.c
thibault-martinez: I think something like this would work but ok, I'll approve.
```
memset(tx->data.signature_or_message, FLEX_TRIT_NULL_VALUE, NUM_FLEX_TRITS_SIGNATURE);
flex_trits_from_trytes(tx->data.signature_or_message, NUM_TRITS_SIGNATURE, (tryte_t *)msg, NUM_TRYTES_SIGNATURE, msg_len);
```

common/model/bundle.c
thibault-martinez: Thanks, I was missing the fact that `tx` is not the same as in the `BUNDLE_FOREACH ` loop

thibault-martinez (approved):

common/model/bundle.c
oopsmonk: yes, it will work, if `signature_or_message` contains FLEX_TRIT_NULLs or transactions in a bundle are properly init, we should make sure it. maybe I think too much.

thibault-martinez: No worries, I'm just writing ideas that come to my mind, they are not all good or too keep :)

oopsmonk: let's make sure transactions are initialized before putting in a bundle.

PR #1192, opened 2019-05-16
Add option -c for conf file path

### LINKED ISSUE #1160: ciri: add -c CLI option for conf file path

howjmay:
use "ciri/conf.yml" as default configure file path.

Fixes #1160


thibault-martinez (changes requested):
This issue is actually a little bit more complex than that.
Configuration variables are initialised like this: Default -> File -> CLI.
So you need to have the configuration file already.

howjmay:
I will try to fix the problems, thanks!

howjmay:
Could I initialized in the order Default -> CLI -> File, and cover the set the unset configures with the value in File?

thibault-martinez:
@HowJMay You can't. Variables both in CLI and File would have the value of the File and we don't want that. I guess you'll have to do a first iteration just to get the value of `-c` from the CLI, use it to do File and then do CLI for real.

thibault-martinez:
Sorry I didn't see that you fixed it.
Please never hesitate to click the re-request review button, it will send us a notification so that we are aware of the changes.
image

thibault-martinez (approved):

thibault-martinez:
Thanks!

howjmay:
@thibault-martinez Thank you for informing. I will use the re-request next time.><

PR #1194, opened 2019-05-18
consensus: rename local snapshot metadata file extension

wusyong:
QoL change of snapshot metadata extension, so it follows as same as IRI and doesn't need to rename every time.

# Test Plan:
CI pass


thibault-martinez:
I'm fine with this. @tsvisabo any reason it was named differently than IRI ?

thibault-martinez (approved):

tsvisabo (approved):

PR #1207, opened 2019-05-24
common: bundle_transactions_bundle_hash, flex_trit_print

oopsmonk:
For a better user experience, this PR implements:
* `bundle_transactions_bundle_hash`: fetch bundle hash.
* `flex_trit_print`: print flex_trit via stdout.

# Test Plan:
CI pass


tsvisabo (changes requested):
Can you also replace current hash retrieval with calling this function?

oopsmonk:
These methods are not used in cclient library but in user applications.
I will apply `flex_trit_print` to `transaction_obj_dump` that will enhance its readability.

common/trinary/flex_trit.c
```diff
@@ -359,3 +360,11 @@ size_t flex_trits_from_bytes(flex_trit_t *to_flex_trits, size_t to_len, const by
 #endif
   return num_trits;
 }
+
```
tsvisabo, line 363: Thanks :)

tsvisabo (approved):

PR #1208, opened 2019-05-25
CClient: Fix missing transaction hash in iota_client_send_trytes API

### LINKED ISSUE #1201: CClient: transaction hash is missing in iota_client_send_trytes API.
**Describe the bug**
The transaction hash is missing in the returned object `out_transactions` when calling `iota_client_send_trytes`.

**To Reproduce**
1. Call `iota_client_send_trytes`
2. using `transaction_obj_dump` to print the returned object, the transaction hash is NULL.

**Expected behavior**
The returned transaction should contain the transaction hash.

jkrvivian:
Close #1201 
Enable `compute_hash` in `transaction_deserialize_from_trits` then we can get hash

# Test Plan:
1. Call `iota_client_send_trytes`
2. Use `transaction_obj_dump` to print the returned object


oopsmonk (changes requested):
Recalculating transaction hash is too expensive since we can get it from transaction trytes.

jkrvivian:
@oopsmonk , 
I think we can't get transaction hash from trytes directly, is there any other strategy to get hash without calculation?

oopsmonk:
> @oopsmonk ,
> I think we can't get transaction hash from trytes directly, is there any other strategy to get hash without calculation?

yeah, you're right, I thought we could get it from `iota_client_attach_to_tangle` but it's not.

jkrvivian:
Then the only way to get hash is to set the flag and recalculate, right?

oopsmonk:
> Then the only way to get hash is to set the flag and recalculate, right?

yes, let's keep it simple.

oopsmonk (approved):

PR #1210, opened 2019-05-26
iota_client_get_inputs enhancement

### LINKED ISSUE #1199: CClient: iota_client_get_inputs enhancement.
**Is your feature request related to a problem? Please describe.**
`iota_client_get_inputs` calls `iota_client_get_new_address` to get an address list in advance then feeds addresses to `iota_client_get_balances`. so it keeps fetching all balances from the list even the balance reached the threshold.

**Describe the solution you'd like**
Stop doing `iota_client_get_balances` if the balance is satisfied.

**Describe alternatives you've considered**
Checking the threshold on each address instead of a list of addresses.

howjmay:
Generate one new address and fetch one address to `iota_client_get_balances`
each time. And end this function `iota_client_get_inputs` once it meets
the threshold (minimun needed balances)

The time cost of  `iota_client_get_balances` and `iota_client_get_new_address` are bascially the same on my laptop, so I choosed to wrap them in the same loop.
close #1199


oopsmonk:
@HowJMay thanks, can you explant how can this modification will fix #1199? I don't see any changes in the result.

howjmay:
> @HowJMay thanks, can you explant how can this modification will fix #1199? I don't see any changes in the result.

I might misunderstand the description of #1199.
I break the loop once the `out_inputs->total_balance` large or equal to minimun required balance. Thus, the loop doesn't need to fetch all the balances in the blanaces list.

And the if statement below can just check out the bool variable `balance_sufficient` to determine if the seed has sufficient balances for transaction

oopsmonk:
> > @HowJMay thanks, can you explant how can this modification will fix #1199? I don't see any changes in the result.
> 
> I might misunderstand the description of #1199.
> I break the loop once the `out_inputs->total_balance` large or equal to minimun required balance. Thus, the loop doesn't need to fetch all the balances in the blanaces list.
> 
> And the if statement below can just check out the bool variable `balance_sufficient` to determine if the seed has sufficient balances for transaction

Okay, the problem is not at the balance checking. 
The problem is `balances_req` containing a list of addresses when calling `iota_client_get_balances` so the API will not stop before fetching all balances in the list.

howjmay:
Thank you for your explanation. May I close this PR, and reopen it once I finish it in a right way?

oopsmonk:
> Thank you for your explanation. May I close this PR, and reopen it once I finish it in a right way?

sure, thank for your help :-)

howjmay:
@oopsmonk I plan to do in the following way
I will fetch 5 addresses each time I call `iota_client_get_balances`
but I need to new/free balances_[req, res] object each time before I call `iota_client_get_balances`.
I am not sure is it a acceptable way to solve this issue?

oopsmonk:
> @oopsmonk I plan to do in the following way
> I will fetch 5 addresses each time I call `iota_client_get_balances`
> but I need to new/free balances_[req, res] object each time before I call `iota_client_get_balances`.
> I am not sure is it a acceptable way to solve this issue?

you can benchmark your solution, then you can figure out it.

howjmay:
@oopsmonk  I have pushed new commit. Is it a more appropriate solution?

cclient/api/extended/get_inputs.c
```diff
@@ -13,51 +13,60 @@
...
   out_inputs->total_balance = 0;
 
   // get address list
-  ret_code = iota_client_get_new_address(serv, seed, addr_opt, &balances_req->addresses);
-  if (ret_code) {
-    log_error(client_extended_logger_id, "%s get new address failed: %s\n", __func__, error_2_string(ret_code));
-    goto done;
-  }
+  address_opt_t addr_opt_elt = {addr_opt.start, 1, addr_opt.security};
+  for (uint64_t addr_index = addr_opt.start; addr_index < addr_opt.total; addr_index++) {
```
oopsmonk: for getting the unused address `addr_opt.total` could be 0.

cclient/api/extended/get_inputs.c
```diff
@@ -13,51 +13,60 @@
...
+    balances_req = get_balances_req_new();
+    balances_res = get_balances_res_new();
+    log_debug(client_extended_logger_id, "[%s:%d]\n", __func__, __LINE__);
+    if (!balances_req || !balances_res) {
+      ret_code = RC_OOM;
+      log_error(client_extended_logger_id, "%s create get balances request or response object failed: %s\n", __func__,
+                error_2_string(ret_code));
+      return ret_code;
```
oopsmonk: memory leak

cclient/api/extended/get_inputs.c
```diff
@@ -13,51 +13,60 @@
...
+    }
+
+    ret_code = iota_client_get_new_address(serv, seed, addr_opt_elt, &balances_req->addresses);
```
oopsmonk: how about using `iota_sign_address_gen_flex_trits` instead of `iota_client_get_new_address`?

oopsmonk (changes requested):

cclient/api/extended/get_inputs.c
howjmay: In this new code, I used it.

cclient/api/extended/get_inputs.c
```diff
@@ -4,63 +4,130 @@
...
 #include "cclient/api/extended/get_inputs.h"
-#include "cclient/api/extended/get_new_address.h"
+#include "cclient/api/core/find_transactions.h"
+#include "cclient/api/core/get_balances.h"
 #include "cclient/api/extended/logger.h"
+#include "common/helpers/sign.h"
 
-retcode_t iota_client_get_inputs(iota_client_service_t const* const serv, flex_trit_t const* const seed,
-                                 address_opt_t const addr_opt, uint64_t const threshold, inputs_t* const out_inputs) {
+static retcode_t is_unused_address(iota_client_service_t const* const serv, flex_trit_t const* const addr,
```
oopsmonk: please call `is_unused_address` in get_new_address.c

cclient/api/extended/get_inputs.c
```diff
@@ -4,63 +4,130 @@
...
-        tmp_input.balance = get_balances_res_balances_at(balances_res, counter);
-        tmp_input.key_index = counter;
+  out_inputs->total_balance = 0;
+
+  flex_trit_t* addr = NULL;
+  bool is_unused = false;
+  uint64_t addr_index, addr_index_end;
+  if (addr_opt.total == 0) {
+    addr_index = 0;
+    addr_index_end = (unsigned)-1;
```
oopsmonk: `addr_index_end` is `uint64_t`

cclient/api/extended/get_inputs.c
```diff
@@ -4,63 +4,130 @@
...
+      ret_code = is_unused_address(serv, addr, &is_unused);
+      if (ret_code == RC_OK && is_unused) {
+        break;
+      }
+    }
+    
+    balances_req->threshold = 100;  // currently `100` should be used
+
+    ret_code = iota_client_get_balances(serv, balances_req, balances_res);
+    if (ret_code == RC_OK && addr) {
```
oopsmonk: `addr` no need to check again.

cclient/api/extended/get_inputs.c
```diff
@@ -4,63 +4,130 @@
...
       }
-      counter++;
-    }
-    // check if balance is sufficient
-    if (out_inputs->total_balance < (int64_t)threshold) {
-      ret_code = RC_CCLIENT_INSUFFICIENT_BALANCE;
-      log_warning(client_extended_logger_id, "insufficient balance\n");
     }
+
+    get_balances_req_free(&balances_req);
```
oopsmonk: you can reuse the request and response instead of allocating every time.

cclient/api/extended/get_inputs.c
```diff
@@ -4,63 +4,130 @@
...
+  }
+
+  // check if balance is sufficient
+  if (out_inputs->total_balance < threshold) {
+    ret_code = RC_CCLIENT_INSUFFICIENT_BALANCE;
+    log_warning(client_extended_logger_id, "insufficient balance\n");
   }
 
 done:
+  free(addr);
```
oopsmonk, line 100: memory leak.

oopsmonk (changes requested):

cclient/api/extended/get_inputs.c
howjmay: but `is_unused_address` is a static function in `get_new_address.c` which is not able to be included by `get_inputs.c`
That's why I use this ugly solution

cclient/api/extended/get_inputs.c
howjmay: I want to assign maximun possible value of `uint64_t` to `addr_index_end`. Use an `(unsigned)-1` would be a simple way without including other header file

cclient/api/extended/get_inputs.c
howjmay: `hash243_queue_to_json_array` would convert all the elements in `hash243_queue_t` to json array,  so if I didn't free the `hash243_queue_t` variable, then `hash243_queue_to_json_array` would generate a wrong json array.
In the other hand, if reuse them without freeing the `get_balances_req_t` and `get_balances_res_t` object, then there will be memory leak problems, since the elements such as `hash243_queue_t` should be freed.

cclient/api/extended/get_inputs.c
howjmay: sorry I used `asan` and examine the code but I don't know which object cause memory leak.

howjmay:
Do we need to consider the situation that a user want to move all the balances to the unused address?

cclient/api/extended/get_inputs.c
oopsmonk: feel free to make `is_unused_address` public in get_new_address.c

cclient/api/extended/get_inputs.c
oopsmonk: the maximum value of `uint64_t`(18446744073709551615) is no equal to `unsigned int -1`(4294967295)

cclient/api/extended/get_inputs.c
oopsmonk: you can clean the queue instead of freeing the entire struct, isn't it?

cclient/api/extended/get_inputs.c
oopsmonk: how did you examine this API call? the `addr` has a memory leak issue from the logic, but I can update you a sanitizer report later.

oopsmonk:
```
==30356==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 324 byte(s) in 4 object(s) allocated from:
	#0 0x7f6e02487108 in calloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xee108)
	#1 0x55d92ab43984 in iota_sign_address_gen_flex_trits common/helpers/sign.c:92
	#2 0x55d92a9fb86f in iota_client_get_inputs cclient/api/extended/get_inputs.c:76
	#3 0x55d92a9e9738 in example_get_inputs cclient/api/examples/example_get_inputs.c:25
	#4 0x55d92a9e4de7 in main cclient/api/examples/cclient_examples.c:88
	#5 0x7f6e01683b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
```

oopsmonk:
> Do we need to consider the situation that a user want to move all the balances to the unused address?

will that be a problem? I don't know what you were referring to?

cclient/api/extended/get_inputs.c
howjmay: oh I got the reason why I didn't get the memory leak issue during testing last time!

cclient/api/extended/get_inputs.c
howjmay: sure

howjmay:
Since if the user want to transfer all the balances into the unused address, then some steps of the current implementation are unnecessary, and we can call both `iota_client_get_new_address` and `iota_client_get_balances` once only to get all the balances we need which can save some time.

oopsmonk:
> Since if the user want to transfer all the balances into the unused address, then some steps of the current implementation are unnecessary, and we can call both `iota_client_get_new_address` and `iota_client_get_balances` once only to get all the balances we need which can save some time.

in general, this API helps to get enough balance from the account(seed), moving all balances could be done in other ways as you mentioned.

howjmay:
> > Since if the user want to transfer all the balances into the unused address, then some steps of the current implementation are unnecessary, and we can call both `iota_client_get_new_address` and `iota_client_get_balances` once only to get all the balances we need which can save some time.
> 
> in general, this API helps to get enough balance from the account(seed), moving all balances could be done in other ways as you mentioned.

Got it. Thank you for clarifying my worries.

oopsmonk (approved):

PR #1213, opened 2019-05-28
mam: improve README

thibault-martinez:


mam/README.md
```diff
@@ -1,60 +1,275 @@
...
-- Message encryption via NTRU (public key encrytpion) and PSK (Pre-Shared Key)
+Masked Authenticated Messaging (MAM) is a second layer data communication protocol which adds functionality to emit and access encrypted data streams, like RSS, over the Tangle. IOTA’s consensus protocol adds integrity to these message streams. Given these properties, MAM fulfills an important need in industries where integrity and privacy meet.
 
-- Sending a message to multiple receivers, each having his/her own session key
+Using MAM you can:
 
-- Authenticating a message either via MAC (Integrity) or via MSS signature (Integrity + Authenticity)
+- create channels for broadcasting messages
+- create channel endpoints for protecting messages during broadcasting
+- protect messages in different ways, for example, turn on / off encryption / authentication
```
tsvisabo: what do you mean by "protect"?

mam/README.md
```diff
@@ -1,60 +1,275 @@
...
+- split messages into parts (packets), protect and transmit each part almost independently
+- set message recipients and provide them with key material in different ways
 
-## MAM components (directories within this repo)
+### Versions
 
-*API* - Translates MAM messages into Tangle's bundle
+A lot of names have been used in the past, either by the Iota Foundation or by the community, to refer to two iterations of the MAM project. You may have heard, for example: MAM prototype, MAM, MAM0, MAM1, MAM2, MAM+...
```
tsvisabo: A lot of -> Many

mam/README.md
```diff
@@ -1,60 +1,275 @@
...
 
-*PSK* - Pre-Shared Key (PSK) is a secret key of Authenticated Encryption (AE). It is preliminarily transmitted between the entities and is beyond the scope of MAM. The PSK id is an identifier of a group of recipients who share the same PSK.
+MAM is a library that read and write IOTA bundles. It must be paired with a client library to actually receive and send them over the Tangle.
 
-*PRNG* - Pseudo randomness generator based on sponge's hash function
+3.4 channel: a source of messages. Belongs to an entity. Identified by a public key, called chid, corresponding to which private keys are used to sign either endpoints (main functionality) or messages;
 
-*PB3* - Encoding implementation used for serialization and by the MAM layer mostly
+3.5 endpoint: a transmitter of messages. Belongs to a channel. Identified by a public key, called epid, corresponding to which private keys are used to sign messages;
+3.6 central endpoint: an endpoint which epid is equal to chid of the corresponding channel;
```
tsvisabo: Should push a new line since in _rich_ layout it seems 3.6 does not start in a new line

mam/README.md
```diff
@@ -1,60 +1,275 @@
...
+- Session keys
+- NTRU private keys
+- Pre-Shared Keys
+- WOTS private keys
+
+Let's say you generated the seed `MAMEXAMPLEMAMEXAMPLEMAMEXAMPLEMAMEXAMPLEMAMEXAMPLEMAMEXAMPLEMAMEXAMPLEMAMEXAMPLE9`.
+
+> **WARNING**: Do not use your regular token seed for MAM !
+
+> **WARNING**: Do not give access to your seed to anyone !
```
tsvisabo: Do not share

mam/README.md
```diff
@@ -1,60 +1,275 @@
...
+```
+
+We are now going to write the first (and only) packet of the message to the same bundle that already contains the header.
+
+We need to provide:
+- the message ID that uniquely identifies the message
+- a payload and its size, they both need to be in trytes. If your data is not in trytes (ASCII, trits, bytes...), you will first need to convert it using one of the functions provided [here](https://github.com/iotaledger/entangled/tree/develop/common/trinary)
+- a checksum method between none, MAC and signature
+- a boolean value to tell if this is an intermediate packet (`false`) or the last packet of the message (`true`). If this is the last packet, all contexts related to this message will be removed on the sender side as well as the receiver side and it won't be possible to write new packets to this message anymore
+- the bundle you wish to write the packet in
```
tsvisabo: the packet into

mam/README.md
```diff
@@ -1,60 +1,275 @@
...
+```c
+retcode_t mam_api_bundle_read(mam_api_t *const api, bundle_transactions_t const *const bundle, tryte_t **const payload, size_t *const payload_size, bool *const is_last_packet);
+```
+
+Now let's switch to the receiver side. We just fetched a bundle from the Tangle and we are ready to read it. It doesn't matter if the bundle contains only a header, a header and a first packet or only a packet, there is only one function to call and we need to provide:
+
+- the bundle
+- a payload that will either be filled with the payload or set to `NULL` if the bundle contained only a header.
+- a payload size that will either be set to the payload size or to `0` if the bundle contained only a header.
+- a boolean that will either be set to `true` if the packet was the last of the message or `false` otherwise. It will tell you if you should continue looking for packets.
```
tsvisabo: remove: "It will tell you if you should continue looking for packets."

tsvisabo (changes requested):

tsvisabo (dismissed):

PR #1236, opened 2019-05-30
utils: fix warnings

### LINKED ISSUE #1222: utils: fix warnings
`bazel build -c opt //utils/...`
and/or
`bazel test -c opt //utils/...`

howjmay:
Declaring variable `buffer_size` and `offset` in `ssize_t`
which solves the warning that comparison between signed and
unsigned number.

Decorate `read` in `/utils/singed_files.c` with `unsigned`
in if comparison to solve the warning that comparison
between signed and unsigned number.

Set timeout in `utils/handles/socket.c`.

Fixes #1222

### DIFF
--- a/utils/handles/socket.c
+++ b/utils/handles/socket.c
@@ -74,7 +74,12 @@ int tls_socket_send(mbedtls_ctx_t *ctx, char const *data, size_t size) {
   return mbedtls_ssl_write(&ctx->ssl, (const unsigned char *)data, size);
 }
 
-int tls_socket_recv(mbedtls_ctx_t *ctx, char *data, size_t size, uint64_t timeout) {
+int tls_socket_recv(mbedtls_ctx_t *ctx, char *data, size_t size, uint32_t timeout) {
+  if (timeout != 0) {
+    mbedtls_ssl_config *ssl_conf = (mbedtls_ssl_config *)malloc(sizeof(mbedtls_ssl_config));
+    memcpy(ssl_conf, ctx->ssl.conf, sizeof(mbedtls_ssl_config));
+    mbedtls_ssl_conf_read_timeout(ssl_conf, timeout);
+  }
   return mbedtls_ssl_read(&ctx->ssl, (unsigned char *)data, size);
 }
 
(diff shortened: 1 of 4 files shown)


thibault-martinez:
Hi @HowJMay , in a turn of events, I happen to have solved this issue in a previous PR (#1235).
I think a saner approach to taking an issue is to post a message in it saying that you would like to do it so that we don't work on it at the same time.

thibault-martinez:
Does that sound fair to you if I take your timeout piece of code and put it in my PR ?

thibault-martinez:
Or maybe just change your PR to only do the timeout change, as you wish.

howjmay:
@thibault-martinez OK! I will reply the issues that I want to take in the future. 
I am OK with your recommendation. Do whatever is easy to you to fix the issue

thibault-martinez:
I will get #1235 merged and you can transform this issue in the timeout thing then.

howjmay:
Got it. Thank you ~~

utils/handles/socket.c
```diff
@@ -74,7 +74,12 @@ int tls_socket_send(mbedtls_ctx_t *ctx, char const *data, size_t size) {
   return mbedtls_ssl_write(&ctx->ssl, (const unsigned char *)data, size);
 }
 
-int tls_socket_recv(mbedtls_ctx_t *ctx, char *data, size_t size, uint64_t timeout) {
+int tls_socket_recv(mbedtls_ctx_t *ctx, char *data, size_t size, uint32_t timeout) {
+  if (timeout != 0) {
+    mbedtls_ssl_config *ssl_conf = (mbedtls_ssl_config *)malloc(sizeof(mbedtls_ssl_config));
```
oopsmonk, line 79: Your implementation is not matched with the doc: https://tls.mbed.org/api/ssl_8h.html#a0d925033b3a46a48b3f8acc1d743af90
please check the usage of `timeout` configuration in the document and make sure timeout is working properly.

oopsmonk (changes requested):

utils/handles/socket.c
howjmay: Thank you for pointing out my error. I will try to fx it up.

PR #1250, opened 2019-06-03
mam/api: api_wrap - remove redundant and weird set

tsvisabo:
obsolete tag is already been reset with previous call to `transaction_reset`
and it also shouldn't be set by `signature_or_message` value which is confusing


mam/api/api.c
```diff
@@ -37,7 +37,6 @@ static void mam_api_bundle_wrap(bundle_transactions_t *const bundle, trit_t cons
   flex_trits_from_trits(buffer, NUM_TRITS_MESSAGE, address, NUM_TRITS_ADDRESS, NUM_TRITS_ADDRESS);
   transaction_set_address(&transaction, buffer);
   transaction_set_value(&transaction, 0);
```
thibault-martinez: We can also remove this

mam/api/api.c
```diff
@@ -505,6 +504,7 @@ retcode_t mam_api_bundle_read(mam_api_t *const api, bundle_transactions_t const
                               size_t *const payload_size, bool *const is_last_packet) {
   retcode_t ret = RC_OK;
   trit_t tag[NUM_TRITS_TAG];
+  trit_t *const msg_id = tag;
```
thibault-martinez, line 506: Why do we introduce this ? Just for the name ? Is that really necessary ?

thibault-martinez (changes requested):

mam/api/api.c
tsvisabo: kind of annoyed me that in one place we use `msg_id` as key for ctx map and in the other we use `tag`,
if you think it's not necessary, i'll accept

thibault-martinez: I don't have a strong opinion on this, I usually go for less variables when that's possible but I understand the idea

tsvisabo: I am also in favor of less, but i prefer consistency/clarity, so this time, i am ok with the cost

thibault-martinez (approved):

PR #1273, opened 2019-06-10
gossip: fix warnings

thibault-martinez:


tsvisabo (approved):

gossip/services/udp_sender.cc
```diff
@@ -37,6 +37,8 @@ bool udp_endpoint_destroy(endpoint_t *const endpoint) {
   boost::asio::ip::udp::endpoint *destination =
       reinterpret_cast<boost::asio::ip::udp::endpoint *>(endpoint->opaque_inetaddr);
   delete destination;
+  endpoint->opaque_inetaddr = NULL;
```
oopsmonk, line 40: it's no effect if you want to set the caller to NULL you have to pass a pointer to that pointer.
`bool udp_endpoint_destroy(endpoint_t *&endpoint)`

thibault-martinez: No it's ok since we are dereferencing from endpoint which is itself a pointer

oopsmonk: my understanding for setting `opaque_inetaddr` to NULL is to prevent the double free issue or dangling pointer in c/c++, but here is not work cause you set the local variable to NULL, not the caller. 
this line is meaningless since you won't use the `opaque_inetaddr` later in `udp_endpoint_destroy`.

thibault-martinez: This is definitely setting the caller variable to NULL because we dereference an endpoint pointer. I do agree it is less critical than for the TCP version but it can still prevent using `udp_send` on a released UDP endpoint.

PR #1283, opened 2019-06-12
mam/api: export prefix to be defined via target 'defines' list

tsvisabo:


mam/defs.h
```diff
@@ -24,6 +24,13 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#if defined(MAM_EXPORT_JS)
```
oopsmonk: use the preprocessor define.
```suggestion
#ifdef __EMSCRIPTEN__
```

mam/api/BUILD
```diff
@@ -31,3 +39,18 @@ map_generate(
...
+# Build command:
+# bazel clean (recommended)
+# bazel build --spawn_strategy=local --define=workspace=$(bazel info workspace)  --define=output_base=$(bazel info output_base)  --config=emscripten //mam:mam.js
+
+emcc_binary(
+    name = "api.js",
+    srcs = [
+        "api.c",
+    ],
+    defines = ["MAM_EXPORT_JS"],
```
oopsmonk: remove

oopsmonk (changes requested):

mam/api/BUILD
thibault-martinez: @tsvisabo Maybe that solve the issue of needing to have it under /mam/api ?

tsvisabo: but how then where would you define `__EMSCRIPTEN__`?

thibault-martinez: I believe @oopsmonk suggested it's defined by emscripten compilers, right ?

mam/api/api.c
```diff
@@ -261,7 +261,7 @@ static retcode_t mam_api_bundle_read_packet_from_msg(mam_msg_read_context_t *ctx
  * Public functions
  */
 
-retcode_t mam_api_init(mam_api_t *const api, tryte_t const *const mam_seed) {
+MAM_EXPORT retcode_t mam_api_init(mam_api_t *const api, tryte_t const *const mam_seed) {
```
thibault-martinez: Remove from C file

thibault-martinez (changes requested):

mam/api/BUILD
tsvisabo: could be, i'll check as soon as my pc is back to normal

oopsmonk: it's predefined in the toolchain.

thibault-martinez (dismissed):

mam/api/BUILD
oopsmonk: normally, compilers will define some flags or macros for platforms and itself that can help to distinguish the environment. we should use it, not define a new one by us. here `__EMSCRIPTEN__` is always be set when compiling app via emscripten. not sure if it's mentioned in the doc tho. :-P

oopsmonk (dismissed):

thibault-martinez (approved):

PR #1292, opened 2019-06-13
infra: update iota.lib.cpp

### LINKED ISSUE #1228: tanglescope: doesn't build on macOS

thibault-martinez:
Fixes #1228

### DIFF
--- a/tools/bazel.rc
+++ b/tools/bazel.rc
@@ -2,11 +2,8 @@ build --copt -W
 build --copt -Wall
 build --copt -Wextra
 
-build --cxxopt -W
-build --cxxopt -Wall
-build --cxxopt -Wextra
-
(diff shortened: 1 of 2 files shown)


tools/bazel.rc
```diff
@@ -2,11 +2,8 @@ build --copt -W
 build --copt -Wall
 build --copt -Wextra
 
-build --cxxopt -W
-build --cxxopt -Wall
```
oopsmonk, line 6: why remove warnings in C++ only?

thibault-martinez: Because `--copt` actually covers C and C++ so they were added twice on C++.
Ref: https://docs.bazel.build/versions/master/user-manual.html#flag--copt

oopsmonk (approved):

PR #1299, opened 2019-06-17
Utils: implementing hashXXX_queue_copy.

oopsmonk:
* utils: implementing hashXXX_queue_copy
* cclient: is_unused_address enhancement.

### DIFF
--- a/cclient/api/extended/get_new_address.c
+++ b/cclient/api/extended/get_new_address.c
@@ -10,8 +10,8 @@
 #include "cclient/api/extended/logger.h"
 #include "common/helpers/sign.h"
 
-static retcode_t were_address_spent_from(iota_client_service_t const* const serv, flex_trit_t const* const addr,
-                                         bool* const is_unused) {
+static retcode_t was_address_spent_from(iota_client_service_t const* const serv, flex_trit_t const* const addr,
+                                        bool* const is_unused) {
   retcode_t ret_code = RC_ERROR;
   // Is it a spent address?
   were_addresses_spent_from_req_t* addr_spent_req = were_addresses_spent_from_req_new();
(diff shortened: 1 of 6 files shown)


cclient/api/extended/get_new_address.c
```diff
@@ -10,11 +10,9 @@
 #include "cclient/api/extended/logger.h"
 #include "common/helpers/sign.h"
 
-retcode_t is_unused_address(iota_client_service_t const* const serv, flex_trit_t const* const addr,
-                            bool* const is_unused) {
+static retcode_t were_address_spent_from(iota_client_service_t const* const serv, flex_trit_t const* const addr,
```
thibault-martinez: Singular is `was_address_spent_from`

cclient/api/extended/get_new_address.c
```diff
@@ -42,7 +40,16 @@ retcode_t is_unused_address(iota_client_service_t const* const serv, flex_trit_t
...
   }
 
+done:
+  were_addresses_spent_from_req_free(&addr_spent_req);
+  were_addresses_spent_from_res_free(&addr_spent_res);
+  return ret_code;
+}
+
+static retcode_t find_transactions_by_address(iota_client_service_t const* const serv, flex_trit_t const* const addr,
+                                              bool* const is_unused, bool get_tx_list, hash243_queue_t* txs) {
```
thibault-martinez, line 50: Do we need the transactions ordered ? Maybe a list/stack instead of a queue ?

utils/containers/hash/hash_queue.c.tpl
```diff
@@ -67,3 +67,26 @@ flex_trit_t *hash{SIZE}_queue_at(hash{SIZE}_queue_t *const queue, size_t index)
   }
   return NULL;
 }
+
+retcode_t hash{SIZE}_queue_copy(hash{SIZE}_queue_t const src, hash{SIZE}_queue_t *dest, size_t size) {
```
thibault-martinez: Usually, `copy` functions takes `dest` as first parameter and `src` as second parameter, so it can be a bit confusing

thibault-martinez (changes requested):

cclient/api/extended/get_new_address.c
oopsmonk: not sure yet, could be changed later.

thibault-martinez (approved):

PR #1303, opened 2019-06-18
gossip: improve processor perfs

thibault-martinez:
Makes the packet queue pop operation to return the entry. That allows to reduce critical section of concurrent context and to remove a lot of copies.


gossip/components/processor.c
```diff
@@ -265,34 +265,30 @@ static void *processor_routine(processor_t *const processor) {
   lock_handle_lock(&lock_cond);
 
   while (processor->running) {
-    if (processor_is_empty(processor)) {
```
tsvisabo, line 271: Does that mean that we can now have a situation where queue is empty it thread is exited immediately?

tsvisabo (changes requested):

gossip/components/processor.c
thibault-martinez: No, the emptiness is now checked here: https://github.com/iotaledger/entangled/pull/1303/files/fd7bb625b97a59d6a9c310e132c82fb00804ef73#diff-40e64e528e65585a1255b012dcd1e3aaR290

tsvisabo (approved):

PR #1308, opened 2019-06-19
gossip: rename in prevision of network refactoring

thibault-martinez:
In prevision of the network refactoring, this PR:
- Renames the `gossip` folder to `node` to better match its content and main component;
- Renames the `components` folder to `pipeline` to better fit the future architecture of the node;
I initially would have done it in the network refactoring PR but it would have made it really hard to review since all files would have been modified by the rename.


oopsmonk:
>     * Renames the `gossip` folder to `node` to better match its content and main component;
As I know, the gossip folder is a general module currently used by cIRI. We can call it `Node` in cIRI or IRI, but rename it to Node is confusing.

>     * Renames the `components` folder to `pipeline` to better fit the future architecture of the node;
>       I initially would have done it in the network refactoring PR but it would have made it really hard to review since all files would have been modified by the rename.
Where can I get more info about the future architecture of the node (or gossip)?

thibault-martinez:
> As I know, the gossip folder is a general module currently used by cIRI. We can call it Node in cIRI or IRI, but rename it to Node is confusing.

The fact is that this folder already does more than just the gossip, it has neighbours management, caching, network... Everything the node does. The main component of this folder is called `node_t` and technically it could be run on its own. We can think of cIRI as just tying `node_t` and `consensus_t` together so it makes sense to me that that the other folder is called `node`.

> Where can I get more info about the future architecture of the node (or gossip)?

https://github.com/iotaledger/iri/pull/1393

oopsmonk:
> > As I know, the gossip folder is a general module currently used by cIRI. We can call it Node in cIRI or IRI, but rename it to Node is confusing.
> 
> The fact is that this folder already does more than just the gossip, it has neighbours management, caching, network... Everything the node does. The main component of this folder is called `node_t` and technically it could be run on its own. We can think of cIRI as just tying `node_t` and `consensus_t` together so it makes sense to me that that the other folder is called `node`.
> 
how about moving them into ciri?  since they depend on ciri.

thibault-martinez:
They don't depend on cIRI, cIRI depends on them!
I have thought about that actually, but technically they could be run on their own, without needing each others.
I'm fine both ways, @tsvisabo opinion on this ?

oopsmonk:
> They don't depend on cIRI, cIRI depends on them!
> I have thought about that actually, but technically they could be run on their own, without needing each others.
> I'm fine both ways, @tsvisabo opinion on this ?

Sorry, I didn't make it clear, I meant the `conf` of consensus and gossip.

thibault-martinez:
@oopsmonk Done

oopsmonk (approved):

PR #1317, opened 2019-06-30
node: clean and improve broadcaster

thibault-martinez:
Switches the broadcaster queue from hash8019 queue to packet queue to:
- Avoid sending back a broadcasted message to its sender (because we now have access to the sender endpoint) resulting in less packet load on the processor;
- Avoid a lot of conversions between bytes and flex trits;

Also cleans the code and add documentation.


thibault-martinez:
Looks like there are issues with the CI though

oopsmonk (approved):

PR #1328, opened 2019-07-05
ciri: check if file exists before closing it

wusyong:
A segment fault encountered when start running cIRI if conf.yml doesn't
exist. Adding file check to fclose resolves this situation.

# Test Plan:
Pass CI


thibault-martinez (approved):
Thanks!

PR #1337, opened 2019-07-14
local_snapshots: transactions pruning

### LINKED ISSUE #1104: Local Snapshots - DB prunning

tsvisabo:
Fixes #1104 

The pruning is done through the service thread, milestone by milestone
There are no jobs and no serialization of pending transactions which weren't pruned due to a shut down
because pruning is sequential and the way to determine which milestone was lastly pruned is by checking the first milestone in the database against metadata file initial milestone index

Run with both:
`local-snapshots-prunning-enabled`
and 
`local-snapshots-enabled`

set to `true`


common/storage/sql/sqlite3/tests/test_sqlite3.c
```diff
@@ -381,6 +394,33 @@ void test_transactions_arrival_time(void) {
   transaction_free(test_tx);
 }
 
+void test_transactions_delete_one_transaction(void) {
```
thibault-martinez: Can we add a test to actually try that batch deletion can delete many txs ?

ciri/consensus/snapshot/snapshots_service.h
```diff
@@ -65,12 +66,13 @@ retcode_t iota_snapshots_service_destroy(snapshots_service_t *const snapshots_se
  * Takes a snapshot and applies it
  *
  * @param snapshots_service The service
+ * @param pruning_manager The pruning_manager
```
thibault-martinez:
```suggestion
 * @param ps The pruning service
```

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.h
```diff
@@ -31,6 +34,7 @@ typedef struct local_snapshots_manager_s {
   milestone_tracker_t const *mt;
   snapshots_service_t *snapshots_service;
   size_t last_snapshot_transactions_count;
+  pruning_service_t ps;
```
thibault-martinez, line 37: Shouldn't it be in the consensus struct with all the other components ?

common/storage/sql/statements.c
```diff
@@ -105,6 +105,9 @@ char *iota_statement_transaction_metadata_clear =
     "UPDATE " TRANSACTION_TABLE_NAME " SET " TRANSACTION_COL_SNAPSHOT_INDEX "=0," TRANSACTION_COL_SOLID
     "=0," TRANSACTION_COL_VALIDITY "=0";
 
+char *iota_transaction_delete_by_hashes =
+    "DELETE FROM " TRANSACTION_TABLE_NAME " WHERE " TRANSACTION_COL_HASH " IN (?)";
```
thibault-martinez: Why is it a `IN(?)` if we do an SQL transaction ? It's either a `WHERE =` with a transaction or an `IN(?)` without transaction but by concatenating all the hashes ?

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
```diff
@@ -124,10 +126,16 @@ retcode_t iota_local_snapshots_manager_init(local_snapshots_manager_t *lsm,
 
   cond_handle_init(&lsm->cond_local_snapshots);
 
+  if (lsm->conf->local_snapshots.pruning_is_enabled) {
+    iota_local_snapshots_pruning_service_init(&lsm->ps, lsm->snapshots_service->snapshots_provider,
```
thibault-martinez: Check return

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
```diff
@@ -124,10 +126,16 @@ retcode_t iota_local_snapshots_manager_init(local_snapshots_manager_t *lsm,
...
+                                              spent_addresses_service, tips_cache, conf);
+  }
+
   return RC_OK;
 }
 
 retcode_t iota_local_snapshots_manager_start(local_snapshots_manager_t *const lsm) {
+  retcode_t ret;
```
thibault-martinez, line 140: New line

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
```diff
@@ -152,6 +164,10 @@ retcode_t iota_local_snapshots_manager_stop(local_snapshots_manager_t *const lsm
     return RC_OK;
   }
 
+  if (lsm->conf->local_snapshots.pruning_is_enabled) {
+    ERR_BIND_RETURN(iota_local_snapshots_pruning_service_stop(&lsm->ps), ret);
```
thibault-martinez: We should not return in case of failure here. It would mean we never get to signal and join and cIRI would be stuck. We should just assign `ret` and return it at the end.

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
```diff
@@ -172,6 +188,10 @@ retcode_t iota_local_snapshots_manager_destroy(local_snapshots_manager_t *const
     return RC_STILL_RUNNING;
   }
 
+  if (lsm->conf->local_snapshots.pruning_is_enabled) {
+    ERR_BIND_RETURN(iota_local_snapshots_pruning_service_destroy(&lsm->ps), ret);
```
thibault-martinez: Same, we shouldn't return because it would mean not releasing all memory.

ciri/consensus/snapshot/local_snapshots/pruning_service.h
```diff
@@ -0,0 +1,97 @@
...
+  lock_handle_t lock_handle;
+  snapshot_t *new_snapshot;
+  spent_addresses_service_t *spent_addresses_service;
+  tips_cache_t *tips_cache;
+} pruning_service_t;
+
+/**
+ * Initializes a pruning service
+ *
+ * @param ps[out] The pruning service
```
thibault-martinez: The `[]` need to be tied to the `@param` to be detected by Doxygen. Like `@param[out]`.

ciri/consensus/snapshot/local_snapshots/pruning_service.h
```diff
@@ -0,0 +1,97 @@
...
+ * @param snapshot_provider[in] The snapshots provider
+ * @param spent_addresses_service[in, out] The spent addresses service
+ * @param conf The consensus's conf [in]
+ *
```
thibault-martinez, line 47: You're missing a parameter

thibault-martinez (changes requested):

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.h
tsvisabo: If there will ever be a need for other consensus components to use pruning we can move it,
but for now its best be encapsulated in here

common/storage/sql/statements.c
tsvisabo: how can you delete multiple rows then?

thibault-martinez: By doing an SQL transaction, that's how we do for batch updating snapshot indexes for example. Here you're not really using the `IN()` feature, you're giving it only one hash at a time.

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
tsvisabo: it won't actually because there' `cond_handle_timedwait` blocking, 
but still, good catch

ciri/consensus/tangle/tangle.h
```diff
@@ -107,6 +107,8 @@ retcode_t iota_tangle_transaction_approvers_count(tangle_t const *const tangle,
 
 retcode_t iota_tangle_transaction_metadata_clear(tangle_t const *const tangle);
 
+retcode_t iota_tangle_transaction_delete(tangle_t const *const tangle, hash243_set_t const hashes);
```
thibault-martinez: Can you make it plural ? `iota_tangle_transactions_delete`

common/storage/storage.h
```diff
@@ -117,6 +117,8 @@ extern retcode_t iota_stor_transaction_find(storage_connection_t const* const co
 
 extern retcode_t iota_stor_transaction_metadata_clear(storage_connection_t const* const connection);
 
+extern retcode_t iota_stor_transaction_delete(storage_connection_t const* const connection, hash243_set_t const hashes);
```
thibault-martinez: Can you make it plural ? `iota_stor_transactions_delete`

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
```diff
@@ -110,7 +110,10 @@ bool iota_local_snapshots_manager_should_take_snapshot(local_snapshots_manager_t
 
 retcode_t iota_local_snapshots_manager_init(local_snapshots_manager_t *lsm,
                                             snapshots_service_t *const snapshots_service,
-                                            iota_consensus_conf_t *const conf, milestone_tracker_t const *const mt) {
+                                            iota_consensus_conf_t *const conf, milestone_tracker_t const *const mt,
+                                            spent_addresses_service_t *const spent_addresses_service,
+                                            tips_cache_t *const tips_cache) {
+  retcode_t ret = RC_OK;
```
thibault-martinez, line 116: New line

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
```diff
@@ -124,10 +127,17 @@ retcode_t iota_local_snapshots_manager_init(local_snapshots_manager_t *lsm,
...
+    ERR_BIND_RETURN(iota_local_snapshots_pruning_service_init(&lsm->ps, lsm->snapshots_service->snapshots_provider,
+                                                              spent_addresses_service, tips_cache, conf),
+                    ret);
+  }
+
+  return ret;
 }
 
 retcode_t iota_local_snapshots_manager_start(local_snapshots_manager_t *const lsm) {
+  retcode_t ret;
```
thibault-martinez, line 140: New line

ciri/consensus/snapshot/local_snapshots/pruning_service.h
```diff
@@ -0,0 +1,98 @@
...
+ * @return a status code
+ */
+retcode_t iota_local_snapshots_pruning_service_destroy(pruning_service_t *const ps);
+
+/**
+ * Updates the last snapshot index so all transactions before it should be pruned
+ *
+ * @param[out] ps The pruning service
+ * @param snapshot[in] The last snapshot
```
thibault-martinez: Tied to `@param`

thibault-martinez (changes requested):

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  *should_stop = false;
+  *should_branch = true;
+  collect_transactions_for_pruning_do_func_params_t *params = data;
+
+  if (pack->num_loaded == 0) {
+    *should_branch = false;
+    return RC_OK;
+  }
+
+  transaction_snapshot_index((iota_transaction_t *)pack->models[0]);
```
thibault-martinez: Why for ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+
+  if (transaction_snapshot_index((iota_transaction_t *)pack->models[0]) == 0) {
+    if ((ret = hash243_set_add(params->transactions_to_prune, hash)) != RC_OK) {
+      return ret;
+    }
+    *should_branch = true;
```
thibault-martinez, line 62: It's already `true` by default

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  return RC_OK;
+}
+
+static retcode_t collect_transactions_for_pruning_do_func(flex_trit_t *hash, iota_stor_pack_t *pack, void *data,
+                                                          bool *should_branch, bool *should_stop) {
+  retcode_t ret = RC_OK;
+  *should_stop = false;
+  *should_branch = false;
```
thibault-martinez, line 72: Maybe we should harmonize the default values

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+
+  if (current_snapshot_index >= params->current_snapshot_index) {
+    *should_branch = true;
+    if ((ret = hash243_set_add(params->transactions_to_prune, hash)) != RC_OK) {
+      return ret;
+    }
+
+    ERR_BIND_RETURN(
+        tangle_traversal_dfs_to_future(params->tangle, collect_unconfirmed_future_transactions_for_pruning_do_func,
+                                       hash, NULL, &params),
```
thibault-martinez: Shouldn't we give `params->transactions_to_prune` as `analyzed_hashes_param` instead of NULL here ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  {
+    connection_config_t db_conf = {.db_path = ps->conf->tangle_db_path};
+
+    if (iota_tangle_init(&tangle, &db_conf) != RC_OK) {
+      log_critical(logger_id, "Failed in initializing db\n");
+      return NULL;
+    }
+  }
+
+  spent_addresses_provider_t sap;
```
thibault-martinez: Can you declare it on top ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  {
+    connection_config_t db_conf = {.db_path = ps->spent_addresses_service->conf->spent_addresses_db_path};
+
+    if (ps->spent_addresses_service->conf->spent_addresses_files == NULL) {
+      return RC_OK;
```
thibault-martinez: You can't return `RC_OK` here.

thibault-martinez: Also, it's fine if there is no file, why return ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    if (iota_spent_addresses_provider_init(&sap, &db_conf) != RC_OK) {
+      log_error(logger_id, "Initializing spent addresses database connection failed\n");
+      goto cleanup;
+    }
+  }
+
+  DECLARE_PACK_SINGLE_MILESTONE(milestone, milestone_ptr, milestone_pack);
```
thibault-martinez: Declare on top

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  iota_tangle_milestone_load_first(&tangle, &milestone_pack);
+
+  lock_handle_lock(&ps->lock_handle);
```
thibault-martinez: The lock doesn't really have to be in the struct, for the other components we declare it directly on the routine

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+      if (prune_transactions(ps, &tangle, &sap) != RC_OK) {
+        goto cleanup;
+      }
+      cond_handle_timedwait(&ps->cond_pruning_service, &ps->lock_handle, PRUNING_RESCAN_INTERVAL_SHORT_MS);
+    }
+    lock_handle_lock(&ps->lock_handle);
+    end_timestamp = current_timestamp_ms();
+    log_info(logger_id, "Pruning from % " PRIu64 " to % " PRIu64 " took % " PRIu64 " milliseconds\n", start_index,
+             ps->last_pruned_snapshot_index, end_timestamp - start_timestamp);
+    cond_handle_timedwait(&ps->cond_pruning_service, &ps->lock_handle, PRUNING_RESCAN_INTERVAL_MS);
```
thibault-martinez: Isn't 1s a bit too much ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  memset(ps, 0, sizeof(pruning_service_t));
+  logger_helper_release(logger_id);
+  lock_handle_destroy(&ps->lock_handle);
+
+  return ret;
+}
+
+void iota_local_snapshots_pruning_service_update_current_snapshot(pruning_service_t *const ps,
+                                                                  snapshot_t *const snapshot) {
+  lock_handle_lock(&ps->lock_handle);
```
thibault-martinez: You shouldn't mix the lock needed by the cond var and the lock needed for the data

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  ps->last_snapshot_index_to_prune = snapshot->metadata.index;
+  lock_handle_unlock(&ps->lock_handle);
+
+  ps->new_snapshot = snapshot;
+  cond_handle_signal(&ps->cond_pruning_service);
```
thibault-martinez, line 351: Why is there a timer in the first place since we can signal it when we need it ? I don't think we need a timer at all.

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+
+retcode_t iota_local_snapshots_pruning_service_start(pruning_service_t *const ps) {
+  if (ps == NULL) {
+    return RC_NULL_PARAM;
+  }
+
+  ps->running = true;
+
+  log_info(logger_id, "Spawning pruning service thread\n");
+  if (thread_handle_create(&ps->pruning_service_thread, (thread_routine_t)pruning_service_routine, ps) != 0) {
```
thibault-martinez, line 296: Is it a thread because we expect the pruning to take a long time ? Can't it be done by the LS manager thread itself ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  ERR_BIND_GOTO(collect_transactions_to_prune(ps, tangle, sap, milestone.hash, &transactions_to_prune), err, cleanup);
+  ERR_BIND_GOTO(iota_tangle_transaction_delete(tangle, transactions_to_prune), err, cleanup);
+  hash243_set_free(&transactions_to_prune);
+  // It's important to delete the milestone only after all it's past cone has been deleted to avoid dangle transactions
+  hash243_set_add(&transactions_to_prune, milestone.hash);
+  ERR_BIND_GOTO(iota_tangle_milestone_delete(tangle, milestone.hash), err, cleanup);
+  log_info(logger_id,
+           "Snapshot index % " PRIu64 " was pruned successfully, % " PRIu64 " transactions were removed from db\n",
+           ps->last_pruned_snapshot_index, hash243_set_size(transactions_to_prune));
+  ps->last_pruned_snapshot_index++;
```
thibault-martinez: Isn't it actually the next snapshot index to be pruned ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  hash243_set_t transactions_to_prune = NULL;
+
+  ERR_BIND_GOTO(iota_tangle_milestone_load_by_index(tangle, ps->last_pruned_snapshot_index, &milestone_pack), err,
+                cleanup);
+
+  ERR_BIND_GOTO(collect_transactions_to_prune(ps, tangle, sap, milestone.hash, &transactions_to_prune), err, cleanup);
+  ERR_BIND_GOTO(iota_tangle_transaction_delete(tangle, transactions_to_prune), err, cleanup);
+  hash243_set_free(&transactions_to_prune);
+  // It's important to delete the milestone only after all it's past cone has been deleted to avoid dangle transactions
+  hash243_set_add(&transactions_to_prune, milestone.hash);
```
thibault-martinez: Why ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  ERR_BIND_GOTO(iota_tangle_milestone_delete(tangle, milestone.hash), err, cleanup);
+  log_info(logger_id,
+           "Snapshot index % " PRIu64 " was pruned successfully, % " PRIu64 " transactions were removed from db\n",
+           ps->last_pruned_snapshot_index, hash243_set_size(transactions_to_prune));
```
thibault-martinez: You won't have the right size since you freed the set. You'll always have `1` here.

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+  hash243_set_entry_t *iter = NULL, *tmp = NULL;
+  bool spent;
+
+  params.current_snapshot_index = ps->last_pruned_snapshot_index + 1;
+
+  ERR_BIND_GOTO(tangle_traversal_dfs_to_past(params.tangle, collect_transactions_for_pruning_do_func, entry_point_hash,
+                                             ps->conf->genesis_hash, NULL, &params),
+                err, cleanup);
+
+  entry_point_has_no_solid_entry_points_in_past_cone = true;
```
thibault-martinez: Init at declaration ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,333 @@
...
+      ERR_BIND_GOTO(
+          iota_tangle_transaction_load_partial(tangle, iter->hash, &tx_pack, PARTIAL_TX_MODEL_ESSENCE_METADATA), err,
+          cleanup);
+      ERR_BIND_GOTO(iota_spent_addresses_service_was_tx_spent_from(sap, tangle, &tx, iter->hash, &spent), err, cleanup);
+      ERR_BIND_GOTO(tips_cache_remove(ps->tips_cache, iter->hash), err, cleanup);
+    }
+  }
+
+  if (!entry_point_has_no_solid_entry_points_in_past_cone) {
+    hash243_set_free(params.transactions_to_prune);
```
thibault-martinez, line 217: Shouldn't it always be freed ? And there is a potential double free with what follows.

thibault-martinez (changes requested):

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: because it's not included in the set and the milestone transaction has to be deleted too, but only after its descendants are removed succesfully

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: whoopsy

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: true

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: No, it should be freed only if there was an error or any of the transactions in the past cone is a SEP,
there is not double free because `entry_point_has_no_solid_entry_points_in_past_cone == false` implies `err == RC_OK`, but i will make it explicit

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: No actually, it's the last index that was pruned

common/storage/storage.h
```diff
@@ -117,6 +117,9 @@ extern retcode_t iota_stor_transaction_find(storage_connection_t const* const co
 
 extern retcode_t iota_stor_transaction_metadata_clear(storage_connection_t const* const connection);
 
+extern retcode_t iota_stor_transaction_delete_transactions(storage_connection_t const* const connection,
```
thibault-martinez: Can you make it `iota_stor_transactions_delete` for consistency ?

ciri/consensus/tangle/tangle.h
```diff
@@ -107,6 +107,8 @@ retcode_t iota_tangle_transaction_approvers_count(tangle_t const *const tangle,
 
 retcode_t iota_tangle_transaction_metadata_clear(tangle_t const *const tangle);
 
+retcode_t iota_tangle_transaction_delete_transactions(tangle_t const *const tangle, hash243_set_t const hashes);
```
thibault-martinez: Can you make it `iota_tangle_transactions_delete` for consistency ?

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
thibault-martinez, line 116: New line

ciri/consensus/snapshot/local_snapshots/local_snapshots_manager.c
thibault-martinez, line 140: New line

ciri/consensus/snapshot/local_snapshots/pruning_service.c
thibault-martinez: @tsvisabo

ciri/consensus/snapshot/local_snapshots/pruning_service.c
thibault-martinez: No but I mean, you are putting `milestone.hash` in the set and then do nothing with the set, why ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,324 @@
...
+  bool spent;
+
+  params.current_snapshot_index = ps->last_pruned_snapshot_index + 1;
+
+  ERR_BIND_GOTO(tangle_traversal_dfs_to_past(params.tangle, collect_transactions_for_pruning_do_func, entry_point_hash,
+                                             ps->conf->genesis_hash, transactions_to_prune, &params),
+                err, cleanup);
+
+  iter = NULL;
+  tmp = NULL;
```
thibault-martinez: They are already initialized

thibault-martinez (changes requested):

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: No, it's the `wait` for when pruning is done, next one should start when LS is done, that could take longer actaully

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: no, it's used from the `local_snapshots_manager` thread and it's protecting `last_snapshot_index_to_prune`

thibault-martinez: You can't use it for both the data and the conditional variable

thibault-martinez: Cond var needs to have its own

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: that data is `last_snapshot_index_to_prune` and it should be protected when the routine is looping

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: because we want to wake up between `snapshot_index` to `snapshot_index + 1` even when we are not signaled to allow updating `last_snapshot_index_to_prune`

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: yes, that's the reason, IRI even has jobs for each `milestone`, `unconfirmed transactions branch`,
so i chose the middle

thibault-martinez: Alright

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,335 @@
...
+      log_critical(logger_id, "Failed in initializing db\n");
+      return NULL;
+    }
+  }
+
+  {
+    connection_config_t db_conf = {.db_path = ps->spent_addresses_service->conf->spent_addresses_db_path};
+
+    if (ps->spent_addresses_service->conf->spent_addresses_files != NULL) {
+      if (iota_spent_addresses_provider_init(&sap, &db_conf) != RC_OK) {
```
thibault-martinez: You always need to initialize the provider, it doesn't depends on the files. The files are for the spent addresses service, not provider.

thibault-martinez (changes requested):

common/storage/storage.h
thibault-martinez: @tsvisabo

ciri/consensus/tangle/tangle.h
thibault-martinez: @tsvisabo

tsvisabo: i would think it is less consistent like that, `iota_tangle_transaction_` is the prefix

thibault-martinez: But we already have some plural suffix `iota_tangle_transactions_` for sets, like `iota_tangle_transactions_update_snapshot_index` or `iota_tangle_transactions_update_solid_state`

tsvisabo: ok ok

tsvisabo: i agree it's nicer like that actually

thibault-martinez (dismissed):

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,362 @@
...
+
+  if (pack->num_loaded == 0) {
+    return RC_OK;
+  }
+
+  uint64_t current_snapshot_index = transaction_snapshot_index((iota_transaction_t *)pack->models[0]);
+
+  if (current_snapshot_index >= params->current_snapshot_index) {
+    *should_branch = true;
+    ERR_BIND_RETURN(
```
thibault-martinez: We should not `ERR_BIND_RETURN` here otherwise the `if` line 89 is never evaluated and we don't set `should_stop` to `true` ?

ciri/consensus/snapshot/local_snapshots/pruning_service.c
```diff
@@ -0,0 +1,362 @@
...
+      if (prune_transactions(ps, &tangle, &sap, &should_wait_for_next_snapshot) != RC_OK) {
+        goto cleanup;
+      }
+      if (should_wait_for_next_snapshot) {
+        break;
+      }
+    }
+    end_timestamp = current_timestamp_ms();
+    if (ps->last_pruned_snapshot_index > start_index) {
+      log_info(logger_id, "Pruning from % " PRIu64 " to % " PRIu64 " took % " PRIu64 " milliseconds\n", start_index,
```
thibault-martinez:
```suggestion
      log_info(logger_id, "Pruning from %" PRIu64 " to %" PRIu64 " took %" PRIu64 " milliseconds\n", start_index,
```

thibault-martinez (changes requested):

ciri/consensus/snapshot/local_snapshots/pruning_service.c
tsvisabo: true

thibault-martinez (approved):

PR #1354, opened 2019-07-19
node: requester refactoring

thibault-martinez:


ciri/consensus/transaction_solidifier/transaction_solidifier.c
```diff
@@ -338,9 +338,6 @@ static retcode_t check_approvee_solid_state(transaction_solidifier_t *const ts,
 
   if (pack.num_loaded == 0) {
     *solid = false;
-    if ((ret = request_transaction(ts->transaction_requester, tangle, approvee)) != RC_OK) {
-      log_error(logger_id, "Requesting missing approvee failed\n");
```
tsvisabo, line 344: why not?

tsvisabo (changes requested):

ciri/consensus/transaction_solidifier/transaction_solidifier.c
thibault-martinez: We now request only on milestone solidification so that we know all the missing transactions actually exist

tsvisabo (approved):

PR #1371, opened 2019-07-22
flex_trits_insert: check for aligned trits to use memcpy

semenov-vladyslav:
Added a special case in `flex_trits_insert` and `flex_trits_insert_from_pos` such that `memcpy` is used when source and destination flex trits are aligned.

# Test Plan:
Added tests (`test_flex_trits_insert`) which cover boundary conditions.
Also, need to run benchmarks and profiling.

### DIFF
--- a/common/trinary/flex_trit.c
+++ b/common/trinary/flex_trit.c
@@ -98,8 +98,17 @@ size_t flex_trits_insert(flex_trit_t *const to_flex_trits, size_t const to_len, 
 #if defined(FLEX_TRIT_ENCODING_1_TRIT_PER_BYTE)
   memcpy(to_flex_trits + start, flex_trits, num_trits);
 #else
-  for (size_t i = 0; i < num_trits; i++) {
-    flex_trits_set_at(to_flex_trits, to_len, start + i, flex_trits_at(flex_trits, len, i));
+  if (0 == (start % NUM_TRITS_PER_FLEX_TRIT)) {
+    // Copy flex trits as bytes
+    memcpy(to_flex_trits + start / NUM_TRITS_PER_FLEX_TRIT, flex_trits, num_trits / NUM_TRITS_PER_FLEX_TRIT);
+    // Handle tail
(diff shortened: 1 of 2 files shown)


common/trinary/flex_trit.c
```diff
@@ -98,8 +98,17 @@ size_t flex_trits_insert(flex_trit_t *const to_flex_trits, size_t const to_len,
 #if defined(FLEX_TRIT_ENCODING_1_TRIT_PER_BYTE)
   memcpy(to_flex_trits + start, flex_trits, num_trits);
 #else
-  for (size_t i = 0; i < num_trits; i++) {
-    flex_trits_set_at(to_flex_trits, to_len, start + i, flex_trits_at(flex_trits, len, i));
+  if (0 == (start % NUM_TRITS_PER_FLEX_TRIT)) {
```
tsvisabo, line 101: Can we not always "handle head" so we get "new" start index and always memcpy from there?

tsvisabo (changes requested):

common/trinary/flex_trit.c
tsvisabo: actually no :(

tsvisabo (approved):

thibault-martinez:
Before we merge I'd be interested in a benchmark, even very basic just
timing the test you wrote with and without your change.

Le mar. 23 juil. 2019 à 08:48, Tsvi Sabo <notifications@github.com> a
écrit :

semenov-vladyslav:
`flex_trits_insert` is the bottleneck in `transaction_serialize`. All the chunks of a transaction are of fixed size and are multiple of 3, which make it perfect for TE3: all the buffers are aligned and `memcpy` is used. It's easy to confirm: 
- 1M serializations of a transaction *without* optimization takes approx 90 seconds, 
- *with* optimization -- a bit more than 1 second.

thibault-martinez (approved):

PR #1379, opened 2019-07-24
flex_trits_slice: use memcpy for aligned buffers

semenov-vladyslav:
Use `memcpy` in `flex_trits_slice` when buffers are aligned, i.e. `start` is on the boundary of `flex_trit_t`.

# Test Plan:
The changes in the code are similar for TE3, TE4, and TE5; the difference is in handling tails.
`test_flex_trits_slice` in tests seem to cover edge cases and succeed for `trit_encoding` equal 3, 4, and 5.
Simple benchmark works as expected: 10M transaction deserializations work in 7 seconds.

### DIFF
--- a/common/trinary/flex_trit.c
+++ b/common/trinary/flex_trit.c
@@ -25,65 +25,101 @@ size_t flex_trits_slice(flex_trit_t *const to_flex_trits, size_t const to_len, f
   // num_bytes == num_trits in a 1:1 scheme
   memcpy(to_flex_trits, flex_trits + start, num_bytes);
 #elif defined(FLEX_TRIT_ENCODING_3_TRITS_PER_BYTE)
-  trit_t trits[6] = {0};
-  size_t index = start / 3U;
-  size_t offset = start % 3U;
-  size_t i, j;
-  size_t end_index = NUM_FLEX_TRITS_FOR_TRITS(len) - 1;
-  for (i = index, j = 0; j < num_bytes; i++, j++) {
-    trytes_to_trits(&flex_trits[i], trits, 1);
-    if (offset && i < end_index) {
-      trytes_to_trits(&flex_trits[i + 1], trits + 3, 1);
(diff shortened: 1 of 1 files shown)


common/trinary/flex_trit.c
```diff
@@ -25,65 +25,101 @@ size_t flex_trits_slice(flex_trit_t *const to_flex_trits, size_t const to_len, f
...
-    buffer = buffer >> tshift;
-    if (rshift && i < end_index) {
-      buffer |= (flex_trits[i + 1] << rshift);
+  if (0 == (start % NUM_TRITS_PER_FLEX_TRIT)) {
+    // Copy flex trits as bytes
+    memcpy(to_flex_trits, flex_trits + start / NUM_TRITS_PER_FLEX_TRIT, num_trits / NUM_TRITS_PER_FLEX_TRIT);
+    // Handle tail
+    if (0 != num_trits % NUM_TRITS_PER_FLEX_TRIT) {
+      uint8_t buffer = 0;
+      uint8_t shift = (4 - num_trits % NUM_TRITS_PER_FLEX_TRIT) << 1U;
```
tsvisabo: this code is identical to that of residual calculation (line 88)
can you create a static function that handle that?

tsvisabo (changes requested):

common/trinary/flex_trit.c
semenov-vladyslav: I don't know a nice way to do that. Something like:
```
flex_trit_t flex_trits_te4_zero_high_trits(flex_trit_t f, size_t n) {
  // n <= 4
  size_t shift = (4 - n) << 1U;
  f <<= shift;
  f >>= shift;
  return f;
}
```

semenov-vladyslav: Actually, I had an idea to refactor/rewrite a bit that code. Maybe try to make it more readable.

tsvisabo (dismissed):

tsvisabo (approved):

PR #1382, opened 2019-07-25
ciri: Add rule to build docker image

### LINKED ISSUE #1353: cIRI: Dockerfile
providing a Dockerfile for running cIRI in a container.

wusyong:
Resolve #1353 

# Test Plan:
Pass CI


ciri/BUILD
```diff
@@ -40,6 +41,11 @@ cc_binary(
     ],
 )
 
+cc_image(
```
oopsmonk, line 45: The `cc_image` contains only ciri and runtime dependencies, not a container, you cannot configure ciri via `cc_image`. 
the optional flags do not work in README.md, in this case, you should set up `container_image` as well.
> bazel run -c opt --define network=mainnet -- ciri:ciri_image # optional flags

oopsmonk (changes requested):

ciri/BUILD
wusyong: After a few testing, `cc_image` should have [distroless](https://console.cloud.google.com/gcr/images/distroless/GLOBAL/cc) image. But it does lack some more info. I will set up `container_image` and give more comprehensive instruction in `README`.

oopsmonk (approved):

PR #1383, opened 2019-07-25
consensus/tipsel: reset boolean to avoid false positive

tsvisabo:


ciri/consensus/tip_selection/exit_probability_randomizer/walker.c
```diff
@@ -67,6 +67,7 @@ static retcode_t find_tail_if_valid(tangle_t *const tangle, exit_prob_transactio
     return ret;
   }
   if (*has_valid_tail) {
+    *has_valid_tail = false;
```
thibault-martinez, line 70: But it's already being reset to `false` on `iota_consensus_exit_prob_transaction_validator_is_valid`

thibault-martinez (changes requested):

ciri/consensus/tip_selection/exit_probability_randomizer/walker.c
tsvisabo: true

PR #1396, opened 2019-08-08
CClient: fixing ESP32 platform issues.

oopsmonk:
* Disable PoW on ESp32.
* Fix overflow issue in `json_array_to_uint64`

### DIFF
--- a/cclient/api/core/attach_to_tangle.c
+++ b/cclient/api/core/attach_to_tangle.c
@@ -34,6 +34,7 @@ retcode_t iota_client_attach_to_tangle(iota_client_service_t const* const servic
   }
 
   if (service == NULL) {
+#if !defined(__XTENSA__)
     log_info(client_core_logger_id, "[%s:%d] local PoW\n", __func__, __LINE__);
     iota_transaction_t tx = {};
     flex_trit_t* array_elt = NULL;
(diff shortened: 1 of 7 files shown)


cclient/api/core/attach_to_tangle.c
```diff
@@ -66,6 +67,10 @@ retcode_t iota_client_attach_to_tangle(iota_client_service_t const* const servic
         goto done;
       }
     }
+#else
+    log_info(client_core_logger_id, "[%s:%d] Local PoW is not supported on ESP32.\n", __func__, __LINE__);
```
jserv: PoW might be absent not only for ESP32 but also for low-end devices such as Arm Cortex-M based SoC. Would it be better if there is a configurable switch and/or runtime info pointing PoW availability?

common/model/transfer.c
```diff
@@ -21,6 +21,10 @@ retcode_t transfer_message_set_trytes(transfer_t* tf, tryte_t const* const tryte
 }
 
 retcode_t transfer_message_set_string(transfer_t* tf, char const* const str) {
+  if (strlen(str) == 0) {
```
jserv: You can simply rewrite as `if (!str) {`, which is shorter and faster.

oopsmonk: this check is both for NULL pointer and string length. cannot use `if (!str) {` in this case.

jserv: Use `if (!str || !*str)` instead.

oopsmonk: don't see a reason for it, but here might misleading and don't need to call `strlen` twice.

cclient/api/core/attach_to_tangle.c
oopsmonk: it should depend on the platform. the user already has the choice of performing remote/local PoW at runtime.

tsvisabo (approved):

PR #1402, opened 2019-08-19
Cppclient ssl

tsvisabo:
- remove some options from bazel.rc to allow build of boringssl
- implement post_ssl

### DIFF
--- a/cppclient/.bazelrc
+++ b/cppclient/.bazelrc
@@ -1,1 +1,0 @@
-build - -copt - w
--- a/cppclient/.bazelrc
+++ b/cppclient/.bazelrc
@@ -1,0 +1,1 @@
+build - -copt - w
--- a/cppclient/api.h
+++ b/cppclient/api.h
@@ -62,10 +62,6 @@ class IotaAPI {
   explicit IotaAPI(bool useSsl) { _useSsl = useSsl; }
   bool isSsl() const { return _useSsl; }
 
-  void setSslPemPath(const std::string& sslPemPath) { _sslPemPath = sslPemPath; }
-
-  const std::string& sslPemPath() const { return _sslPemPath; }
-
   virtual bool isNodeSolid() = 0;
 
   virtual nonstd::optional<std::unordered_map<std::string, uint64_t>> getBalances(
--- a/cppclient/api.h
+++ b/cppclient/api.h
@@ -59,6 +59,13 @@ using Bundle = std::vector<Transaction>;
 /// - broadcastTransactions
 class IotaAPI {
  public:
+  explicit IotaAPI(bool useSsl) { _useSsl = useSsl; }
+  bool isSsl() const { return _useSsl; }
+
+  void setSslPemPath(const std::string& sslPemPath) { _sslPemPath = sslPemPath; }
+
+  const std::string& sslPemPath() const { return _sslPemPath; }
+
   virtual bool isNodeSolid() = 0;
 
   virtual nonstd::optional<std::unordered_map<std::string, uint64_t>> getBalances(
@@ -98,6 +105,10 @@ class IotaAPI {
 
   virtual nonstd::optional<WereAddressesSpentFromResponse> wereAddressesSpentFrom(
       const std::vector<std::string>& addresses) = 0;
+
+ private:
+  bool _useSsl;
+  std::string _sslPemPath;
 };
 }  // namespace cppclient
 
--- a/cppclient/beast.cc
+++ b/cppclient/beast.cc
@@ -34,9 +34,9 @@ nonstd::optional<json> BeastIotaAPI::post(const json& input) {
 
   try {
     if (isSsl()) {
-      result = post_ssl(input);
+      result = postSsl(input);
     } else {
-      result = post_plain(input);
+      result = postPlainText(input);
     }
   } catch (const std::exception& ex) {
     LOG(ERROR) << ex.what();
@@ -46,7 +46,7 @@ nonstd::optional<json> BeastIotaAPI::post(const json& input) {
   return result;
 }
 
-nonstd::optional<json> BeastIotaAPI::post_plain(const json& input) {
+nonstd::optional<json> BeastIotaAPI::postPlainText(const json& input) {
   net::io_context ioc;
   boost::system::error_code ec;
   tcp::resolver resolver{ioc};
@@ -83,7 +83,7 @@ nonstd::optional<json> BeastIotaAPI::post_plain(const json& input) {
   return result;
 }
 
-nonstd::optional<json> BeastIotaAPI::post_ssl(const json& input) {
+nonstd::optional<json> BeastIotaAPI::postSsl(const json& input) {
   net::io_context ioc;
   boost::system::error_code ec;
   tcp::resolver resolver{ioc};
--- a/cppclient/beast.cc
+++ b/cppclient/beast.cc
@@ -15,7 +15,6 @@
 #include <boost/asio/ssl/error.hpp>
 #include <boost/asio/ssl/stream.hpp>
 #include <boost/beast/core.hpp>
-#include <boost/beast/http.hpp>
 #include <boost/beast/ssl.hpp>
 #include <boost/beast/version.hpp>
 #include <nlohmann/json.hpp>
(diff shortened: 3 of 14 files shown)


cppclient/beast.h
```diff
@@ -24,13 +25,16 @@ namespace cppclient {
 class BeastIotaAPI : virtual public IotaAPI, public IotaJsonAPI {
  public:
   BeastIotaAPI() = delete;
-  BeastIotaAPI(std::string host, uint32_t port) : _host(std::move(host)), _port(port) {}
+  BeastIotaAPI(std::string host, uint32_t port, bool useSsl) : IotaAPI(useSsl), _host(std::move(host)), _port(port) {}
```
oopsmonk, line 28: snake_case

tools/bazel.rc
```diff
@@ -1,7 +1,3 @@
-build --copt -W
-build --copt -Wall
```
oopsmonk, line 2: @thibault-martinez I'm okay with it, do you agree to remove these warning options?

cppclient/beast.cc
```diff
@@ -67,4 +82,70 @@ nonstd::optional<json> BeastIotaAPI::post(const json& input) {
...
+    if (ec && ec != boost::system::errc::not_connected) throw boost::system::system_error{ec};
+  } catch (const std::exception& ex) {
+    LOG(ERROR) << ex.what();
+    return {};
+  }
+
+  return result;
+}
+
+http::request<http::string_body> BeastIotaAPI::setupRequest(const nlohmann::json& input) const {
```
oopsmonk, line 139: snake_case

cppclient/api.h
```diff
@@ -59,6 +59,13 @@ using Bundle = std::vector<Transaction>;
 /// - broadcastTransactions
 class IotaAPI {
  public:
+  explicit IotaAPI(bool useSsl) { _useSsl = useSsl; }
+  bool isSsl() const { return _useSsl; }
```
oopsmonk, line 63: snake_case

cppclient/api.h
```diff
@@ -59,6 +59,13 @@ using Bundle = std::vector<Transaction>;
...
+
+  void setSslPemPath(const std::string& sslPemPath) { _sslPemPath = sslPemPath; }
```
oopsmonk: snake_case

cppclient/api.h
```diff
@@ -59,6 +59,13 @@ using Bundle = std::vector<Transaction>;
...
+
+  const std::string& sslPemPath() const { return _sslPemPath; }
```
oopsmonk: snake_case

cppclient/tests/beast.cc
```diff
@@ -28,11 +28,12 @@ class MockBeast : public BeastIotaAPI {
  public:
   using BeastIotaAPI::BeastIotaAPI;
   using BeastIotaAPI::post;
+  explicit MockBeast(std::string host, uint32_t port, bool useSsl) : IotaAPI(false), BeastIotaAPI(host, port, false) {}
 };
 
 TEST_F(BeastIotaAPITest, InvalidHostPort) {
```
oopsmonk, line 34: please test http/https connections with IRI node as well.

oopsmonk (changes requested):

cppclient/beast.h
tsvisabo: This library was written with camelCase

cppclient/tests/beast.cc
tsvisabo: sure, will get a neighbor today

cppclient/beast.h
oopsmonk: I was mean the `useSsl`, if the style is camelCase then should try to keep the consistency, in this case, `post_plain`, `post_ssl` need a change.

tsvisabo: correct, i overlooked that :)

tools/bazel.rc
thibault-martinez: @tsvisabo Why are we removing them ?

thibault-martinez: I don't have any third party related warnings, I don't think we should remove them.

thibault-martinez:
Not added by your PR but if you could fix the following warning:

```
INFO: From Compiling cppclient/beast.cc:
cppclient/beast.cc:147:10: warning: moving a local object in a return statement prevents copy elision [-Wpessimizing-move]
  return std::move(req);
         ^
cppclient/beast.cc:147:10: note: remove std::move call here
  return std::move(req);
         ^~~~~~~~~~   ~
1 warning generated.

```

tools/bazel.rc
tsvisabo: Because boringssl though external fails with these flags due to unused params
@thibault-martinez has found that boringssl define `-Werror` and it collides with our `-Wextra`
so one solution is of course to remove those flags, another one is to try and suppress flags from dependency though i don't see how

oopsmonk (dismissed):

cppclient/.bazelrc
```diff
@@ -0,0 +1 @@
+build - -copt - w
```
thibault-martinez: What is it for ?
And shouldn't it be `build --copt -w` ?

tsvisabo: it's not used because it's not imported, good catch, i will remove

thibault-martinez (dismissed):

tools/bazel.rc
oopsmonk: I did some test on this PR. it's not related to `boringssl`'s warning options, these compile options only apply to itself, it won't affect the whole project. 
we can keep `-Wall -Wextra`, remove `-W` which is redundant and `-Wno-error=unused-parameter -Wno-error=cast-function-type` they are cause compiling failed on `googletest`, `curl-p`, `trinary` and so on in my env. 
adding these options causes incompatible issues between compilers. we should support different compilers, IMO.

thibault-martinez: > it's not related to boringssl's warning options, these compile options only apply to itself, it won't affect the whole project.

Yes but it's the other way around. We are providing `-Wextra` that is propagated to BoringSSL which has `-Werror` hence the failure. Doing what you say still ends up in compilation error.

thibault-martinez: And how can `googletest`, `curl-p` and `trinary` fail if we don't have a `-Werror` on our end ? Can you send the actual error ? Or do you mean that `-Wno-error=unused-parameter` and  `-Wno-error=cast-function-type` simply doesn't exist on another compiler ?

oopsmonk: > And how can `googletest`, `curl-p` and `trinary` fail if we don't have a `-Werror` on our end ? Can you send the actual error ? Or do you mean that `-Wno-error=unused-parameter` and `-Wno-error=cast-function-type` simply doesn't exist on another compiler ?

gcc 7.4.0 has no `cast-function-type` option.

```
ERROR: /home/sam/.cache/bazel/_bazel_sam/6033e706b65718d7c9e2770fdaa44559/external/com_google_googletest/BUILD.bazel:55:1: C++ compilation of rule '@com_google_googletest//$
gtest' failed (Exit 1) gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-o$
it-frame-pointer '-std=c++0x' -MD -MF ... (remaining 38 argument(s) skipped)

cc1plus: error: -Werror=cast-function-type: no option -Wcast-function-type
```

oopsmonk: > > it's not related to boringssl's warning options, these compile options only apply to itself, it won't affect the whole project.
> 
> Yes but it's the other way around. We are providing `-Wextra` that is propagated to BoringSSL which has `-Werror` hence the failure. Doing what you say still ends up in compilation error.

I wasn't thinking clearly in the middle of the night. :P
alright, `-Wextra` is apply to boringssl, so we just remove the `-Werror` in boringssl.
https://github.com/iotaledger/rules_iota/pull/46

oopsmonk (dismissed):

oopsmonk (approved):

PR #1412, opened 2019-08-27
Pcurl improvements

semenov-vladyslav:
Ptrit/pcurl implementation using SIMD (see [README](https://github.com/semenov-vladyslav/curl/blob/master/ptrit_curl/README.md)).

When building with `bazel` ptrit and pcurl configuration flags must be passed via `--define`:

- **PTRIT_PLATFORM**:
```
bazel ... --define ptrit_platform=PTRIT_64
bazel ... --define ptrit_platform=PTRIT_SSE2 --copt=-msse2
bazel ... --define ptrit_platform=PTRIT_AVX2 --copt=-mavx2
bazel ... --define ptrit_platform=PTRIT_AVX512 --copt=-mavx512f --copt=-fno-asynchronous-unwind-tables
bazel ... --define ptrit_platform=PTRIT_NEON --copt=-march=armv7-a --copt=-mfpu=neon
```

- **PTRIT_CVT**:
```
bazel ... --define ptrit_cvt=PTRIT_CVT_ANDN
bazel ... --define ptrit_cvt=PTRIT_CVT_ORN
```

- **PCURL_S2_CIRCUIT**:
```
bazel ... --define pcurl_s2_circuit=PCURL_S2_CIRCUIT4
bazel ... --define pcurl_s2_circuit=PCURL_S2_CIRCUIT5
```

- **PCURL_S2_ARGS**:
```
bazel ... --define pcurl_s2_args=PCURL_S2_ARGS_PTR
bazel ... --define pcurl_s2_args=PCURL_S2_ARGS_LUT
```

- **PCURL_STATE**:
```
bazel ... --define pcurl_state=PCURL_STATE_SHORT
bazel ... --define pcurl_state=PCURL_STATE_DOUBLE
```

- **PCURL_SBOX_UNWIND**:
```
bazel ... --define pcurl_sbox_unwind=PCURL_SBOX_UNWIND_1
bazel ... --define pcurl_sbox_unwind=PCURL_SBOX_UNWIND_2
bazel ... --define pcurl_sbox_unwind=PCURL_SBOX_UNWIND_4
bazel ... --define pcurl_sbox_unwind=PCURL_SBOX_UNWIND_8
```

# TODO:

- [x] Detect BMI instruction in cpuid, needed in order to use `_tzcnt_u64` intrinsic under MSVC.
- [x] In `ptrits_find_zero_slice` return `SIZE_MAX` instead of `PTRIT_SIZE`?
- [x] Check for carry after `ptrit_increment` in pearl diver, carry can indicate exhaustion of the search.
- [x] The original end of `ptrit_increment` in `run_search_thread` was `HASH_LENGTH_TRIT`. Was it an error? It should be `end`.
- [x] There is no need for `statusLock` in pearl diver. It locks access to `status` variable, which is only overwritten with `SEARCH_FINISHED`. As a compromise atomics can be used instead of lock.
- [x] Search state is incremented in pearl diver as follows: `ptrit_increment(state, begin + 1, end)`. With multiple threads (4+) this search space can overlap?!
- [x] Test on ARM with NEON and on AVX512 under sde.
- [x] `ssize_t` is not standard type and causes problems on Windows with msvc.
- [x] The bazel command line is too long, add PTRIT_PLATFORM autodetect.
- [x] Drop support for PCURL_S2_ARGS_LUT as it won't be used in production code?
- [ ] Drop support for PCURL_SBOX_UNWIND?
- [x] Ensure the ciri uses PTRIT_SIZE transactions per call to pcurl.


semenov-vladyslav:
Also, `ssize_t` causes problems on Windows, as it's not a standard type.

common/trinary/ptrit.h
```diff
@@ -9,15 +9,199 @@
...
+#define XORAND(x, y, z) XOR(x, AND(y, z))
+#define XORORN(x, y, z) XOR(x, ORN(y, z))
+#endif  // PTRIT_AVX512F
+#endif  // PTRIT_NEON
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// The lowest 2 bits of `t` are used.
```
tsvisabo: Can you use this kind of documentation? 
e.g:

/**
 * Add a trusted channel id into the api's trusted channels set
 *
 * @param api - The API [in,out]
 * @param pk - A new public key [in]
 *
 * @return return code
 */

common/trinary/ptrit.h
```diff
@@ -9,15 +9,199 @@
...
+#else
+#define AND(x, y) ((x) & (y))
+#define XOR(x, y) ((x) ^ (y))
+#define OR(x, y) ((x) | (y))
+#define NOT(x) (~(x))
+#define ANDN(x, y) AND(NOT(x), (y))
+#define ORN(x, y) OR(NOT(x), (y))
+#endif  // PTRIT_PLATFORM generic
+
+#define XORANDN(x, y, z) XOR(x, ANDN(y, z))
```
tsvisabo, line 181: wouldn't that override previous definitions?  e.g: `PTRIT_AVX512F` 's `XORANDN`

common/trinary/ptrit.c
```diff
@@ -0,0 +1,159 @@
...
+// TODO: check for cpuid BMI1
+#define CTZLL(x) _tzcnt_u64(x)
+#else
+#define CTZLL(x) __builtin_ctzll(x)
+#endif
+
+// Not-a-Trit
+#define NaT 2
+
+static uint64_t const cvt_te1_to_tep[4][2] = {
```
tsvisabo, line 27: what's cvt? convert?

common/trinary/ptrit.c
```diff
@@ -0,0 +1,159 @@
...
+    {-1, 0},
+    {NaT, +1},
+#else
+#error Invalid PTRIT_CVT.
+#endif  // PTRIT_CVT
+};
+
+void ptrit_fill(ptrit_t *p, trit_t t) {
+  // `t & 3` in order to avoid buffer overflow
+  uint64_t const *c = cvt_te1_to_tep[t & 3];
```
tsvisabo, line 61: not sure we should try to avoid buffer overflow, if trit is not within {-1,0,1} then we would not won't the `ptrit_fill` to be silent, probably should assert here

common/trinary/tests/test_ptrits.c
```diff
@@ -0,0 +1,44 @@
...
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include <unity/unity.h>
+
+#include "common/trinary/ptrit.h"
+#include "common/trinary/ptrit_incr.h"
+
+int trits_to_int(size_t n, trit_t const *t) {
+  int s = 0;
```
tsvisabo: better to use:
https://github.com/iotaledger/entangled/blob/0817986f24417c2931fc2ad6cf4f0a69daccdef1/common/trinary/trit_long.h#L25

(Also, this function is a good candidate for refactoring ;))

common/trinary/ptrit_incr.h
```diff
@@ -14,8 +14,22 @@ extern "C" {
...
+#define PTRIT_SIZE_LOG3 5
+#elif (256 == PTRIT_SIZE)
+#define PTRIT_SIZE_LOG3 6
+#elif (512 == PTRIT_SIZE)
+#define PTRIT_SIZE_LOG3 6
+#else
+#error Invalid PTRIT_SIZE.
+#endif  // PTRIT_SIZE
+
+void ptrits_set_iota(ptrit_t range[PTRIT_SIZE_LOG3]);
```
tsvisabo: documentation please

tsvisabo (changes requested):

common/crypto/curl-p/ptrit.c
```diff
@@ -1,75 +1,433 @@
...
+    pcurl_s2(s + curl_index[i], s + curl_index[i + 1], c + i);
+    ++i;
+    pcurl_s2(s + curl_index[i], s + curl_index[i + 1], c + i);
+    ++i;
+  }
+#elif defined(PCURL_SBOX_UNWIND_8)
+  pcurl_s2(s + curl_index[0], s + curl_index[1], c);
+  for (i = 1; i < CURL_STATE_SIZE;) {
+    pcurl_s2(s + curl_index[i], s + curl_index[i + 1], c + i);
+    ++i;
```
tsvisabo: can you export this code into another loop and set number of iterations based on #define PCURL_SBOX_* value?

common/crypto/curl-p/ptrit.c
```diff
@@ -1,75 +1,433 @@
...
-    c[i].high = (alpha ^ s[CURL_INDEX[i + 1]].high) | delta;
+  // 0, 364, 728, 363, 727, ..., 2, 366, 1, 365, 0
+  ptrit_t const *x = s + 0, *y = s + 364;
+  pcurl_s2(x, y, c++);
+  x = s + 728;
+
+#if defined(PCURL_SBOX_UNWIND_2)
+  // 728 = 8*91
+  for (i = 0; i < CURL_STATE_SIZE / 2; ++i) {
+    pcurl_s2(y--, x, c++);
```
tsvisabo: ditto

tsvisabo (changes requested):

semenov-vladyslav:
Changed pearl diver search logic a bit: search range is split into fixed and variable ranges. The fixed range is initialized with different `n * PTRIT_SIZE` integer values, where `n` is the number of threads. The variable range is initialized with "all -1s", so when `ptrit_hincr` overflows that means that the variable range is completely exhausted and the search has failed.

common/trinary/ptrit.c
semenov-vladyslav: Conversion table

common/trinary/ptrit.c
semenov-vladyslav: Assert wouldn't work in release, and checking all trits for correct values can be too slow. So the idea here and in flex trit conversions is to assign some default value to incorrect input trit (TE1, TE3, TE4, TE5) values. All validity checks must be cryptographic (HASH, MAC, Sig), so checksum verifications in these cases would fail.

common/trinary/tests/test_ptrits.c
semenov-vladyslav: Ok, didn't notice it

common/crypto/curl-p/ptrit.c
semenov-vladyslav: The whole idea was to unwind these loops. Of course we can put them into an internal loop, but will the compiler figure out that this loop needs to be unwound?

common/trinary/ptrit.h
semenov-vladyslav: No, `PTRIT_AVX512F`'s `XORANDN` is in a different `#if` branch.

common/trinary/ptrit.c
tsvisabo: fair enough

common/crypto/curl-p/ptrit.c
tsvisabo: My bad then, 
but seems like there is a gcc option:
https://stackoverflow.com/questions/4071690/tell-gcc-to-specifically-unroll-a-loop/14251545

common/trinary/ptrit_incr.h
semenov-vladyslav: Done

common/crypto/curl-p/ptrit.c
semenov-vladyslav: I'm not sure how to use `__attribute__((optimize("unroll-loops")))`. Does it unroll the loop completely? But I want to unroll just a few rounds. Plus it's gcc-specific. IMO it's easier to just drop support for PCURL_SBOX_UNWIND, the speedup gain seems to be very small - a few percents.

common/trinary/trit_ptrit.c
```diff
@@ -8,60 +8,13 @@
 #include "common/trinary/trit_ptrit.h"
 
 void trits_to_ptrits(trit_t const *const trits, ptrit_t *const ptrits, size_t const index, size_t const length) {
```
thibault-martinez: May be worth to make it `static inline` in the header

common/trinary/trit_ptrit.c
```diff
@@ -8,60 +8,13 @@
...
-        break;
-      default:
-        ptrits[j].low |= (1uLL << index);
-        break;
-    }
-  }
+  ptrits_set_slice(length, ptrits, index, trits);
 }
 
 void trits_to_ptrits_fill(trit_t const *const trits, ptrit_t *const ptrits, size_t const length) {
```
thibault-martinez: May be worth to make it `static inline` in the header

common/trinary/trit_ptrit.c
```diff
@@ -8,60 +8,13 @@
...
-      break;
-    default:
-      ptrits->low = HIGH_BITS;
-      ptrits->high = LOW_BITS;
-  }
-  trits_to_ptrits_fill(&trits[1], &ptrits[1], length - 1);
+  ptrits_fill(length, ptrits, trits);
 }
 
 void ptrits_to_trits(ptrit_t const *const ptrits, trit_t *const trits, size_t const index, size_t const length) {
```
thibault-martinez: May be worth to make it `static inline` in the header

thibault-martinez: And then we don't  even use this file anymore

common/trinary/tests/test_ptrits.c
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2019 IOTA Stiftung
+ * https://github.com/iotaledger/entangled
+ *
+ * Refer to the LICENSE file for licensing information
+ */
+
+#include <string.h>
+#include <unity/unity.h>
```
thibault-martinez, line 10:
```suggestion

#include <unity/unity.h>
```

thibault-martinez: This is my OCD talking, we usually separate headers in 3 sections: standard, libraries, entangled

common/crypto/curl-p/BUILD
```diff
@@ -29,12 +29,81 @@ cc_library(
...
+config_setting(
+    name = "pcurl_sbox_unwind_4",
+    values = {"define": "pcurl_sbox_unwind=PCURL_SBOX_UNWIND_4"},
+)
+
+config_setting(
+    name = "pcurl_sbox_unwind_8",
+    values = {"define": "pcurl_sbox_unwind=PCURL_SBOX_UNWIND_8"},
+)
+
```
thibault-martinez, line 61: For all these settings, we don't really need to put the actual define after the `=` so we could have "simpler" names e.g `pcurl_sbox_unwind=1`, `pcurl_sbox_unwind=2` ... or `pcurl_s2_args=lut`, `pcurl_s2_args=ptr`...

common/trinary/BUILD
```diff
@@ -24,16 +24,65 @@ cc_library(
...
+)
+
+config_setting(
+    name = "ptrit_cvt_orn",
+    values = {"define": "ptrit_cvt=PTRIT_CVT_ORN"},
+)
+
+config_setting(
+    name = "ptrit_cvt_andn",
+    values = {"define": "ptrit_cvt=PTRIT_CVT_ANDN"},
```
thibault-martinez: Same goes for these

common/trinary/ptrit.c
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+#include <assert.h>
+#include <string.h>
+
+#include "common/trinary/ptrit.h"
+
+#if defined(_MSC_VER)
+#include <immintrin.h>
+// TODO: check for cpuid BMI1
```
thibault-martinez: Status on this ?

common/trinary/ptrit.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+void ptrits_get_slice(size_t n, trit_t *dst, ptrit_t const *src, size_t idx) {
+  for (; n--;) {
+    *dst++ = ptrit_get(src++, idx);
+  }
+}
+
+// Return index of all-zero n-trit slice or PTRIT_SIZE.
+// TODO: return SIZE_MAX?
```
thibault-martinez: Status on that ?

thibault-martinez (changes requested):

common/crypto/curl-p/BUILD
semenov-vladyslav: Right, makes sense. Just wanted to make the config names in check with defines names with the hope that they can be reused one day.

common/trinary/ptrit.c
semenov-vladyslav: `_tzcnt_u64` is defined in `<immintrin.h>` as follows:
```
#if defined (_M_X64)
extern unsigned __int64 _tzcnt_u64(unsigned __int64);
#endif  /* defined (_M_X64) */
```

It seems this intrinsic is always defined for x64 targets, I can reuse this condition in the source. But what about 32-bit builds?

common/trinary/ptrit.c
semenov-vladyslav: `SIZE_MAX` is commonly known/recognized, the use of `PTRIT_SIZE` might not be that obvious but a bit more logical. What's the best?

common/trinary/trit_ptrit.c
semenov-vladyslav: Done

common/trinary/trit_ptrit.c
semenov-vladyslav: Done

common/trinary/trit_ptrit.c
semenov-vladyslav: Done

common/trinary/tests/test_ptrits.c
semenov-vladyslav: Done

common/trinary/ptrit.c
semenov-vladyslav: For 32-bit target `_tzcnt_u32` is used.

common/crypto/curl-p/BUILD
semenov-vladyslav: Done

common/trinary/BUILD
semenov-vladyslav: Done

thibault-martinez (changes requested):
I get errors when I try to run tests with `bazel test -c opt //common/...`

```
common/trinary/tests/test_trit_ptrit.c:18:20: error: use of undeclared identifier 'HIGH_BITS'
  ptrit_t exp[] = {ptrit_EXP};
```

thibault-martinez:
I allowed myself to tick two items of your list because I created the following related issues:
- https://github.com/iotaledger/entangled/issues/1422
- https://github.com/iotaledger/entangled/issues/1414

common/crypto/curl-p/hamming.c
```diff
@@ -1,47 +1,40 @@
...
-        } else if ((curl->state[k].high & (1uLL << i)) == 0) {
-          sum++;
-        }
-      }
+static test_result_t test(pcurl_t const *pcurl, test_arg_t security) {
+  size_t i;
+
+  for (i = 0; i < PTRIT_SIZE; i++) {
+    ptrit_t const *p = pcurl->state;
+    ssize_t sum = 0;
```
thibault-martinez: Let's not introduce a new `ssize_t` ?

thibault-martinez (changes requested):

common/crypto/curl-p/ptrit.c
```diff
@@ -1,75 +1,323 @@
...
-      rhs = s.state;
-    } else {
-      lhs = s.state;
-      rhs = ctx->state;
-    }
-    ptrit_sbox(lhs, rhs);
+  ptrit_t *a = ctx->state;
+  ptrit_t *b = a + (CURL_STATE_SIZE + 1) / 2;
+  ptrit_t *c = b + (CURL_STATE_SIZE + 1) / 2;
+  b[364] = a[0];
```
thibault-martinez: Should we use a constant here and everywhere we use this ?

common/trinary/ptrit.c
thibault-martinez: I'd go for the most logical way. Make sure it's properly documented and it should be fine :)

thibault-martinez (changes requested):

common/trinary/ptrit.c
semenov-vladyslav: Done.

common/crypto/curl-p/hamming.c
semenov-vladyslav: Signed `size_t` was a good candidate to hold a "sum of trits". I guess `long` will also do.

semenov-vladyslav: `long` it is. Done.

common/crypto/curl-p/ptrit.c
semenov-vladyslav: Could you please clarify? Not sure what you mean.

semenov-vladyslav: `#define CURL_STATE_HALF_SIZE ((CURL_STATE_SIZE + 1) / 2)`
364 = `CURL_STATE_HALF_SIZE - 1`
Done.

thibault-martinez (dismissed):

thibault-martinez (dismissed):

tsvisabo (dismissed):

thibault-martinez (dismissed):

CMakeLists.txt
```diff
@@ -25,6 +25,8 @@ find_package(Threads REQUIRED)
 # default CLI options
 option(CCLIENT "Enable CClient library" ON)
 option(CCLIENT_TEST "Enable CClient library test cases" ON)
+option(PCURL_SBOX_UNWIND "PCURL_SBOX_UNWIND_1 PCURL_SBOX_UNWIND_2 PCURL_SBOX_UNWIND_4 PCURL_SBOX_UNWIND_8" PCURL_SBOX_UNWIND_2)
```
oopsmonk: cmake option is a boolean value(ON/OFF), it does not work in this way.

CMakeLists.txt
```diff
@@ -25,6 +25,8 @@ find_package(Threads REQUIRED)
...
+option(PCURL_STATE "PCURL_STATE_SHORT PCURL_STATE_DOUBLE" PCURL_STATE_SHORT)
```
oopsmonk: ditto

CMakeLists.txt
```diff
@@ -160,6 +164,23 @@ set(UTILS_SRC
   "${HASH_CONTAINERS_DIR}/hash8019_stack.c"
 )
 
+if(${PCURL_SBOX_UNWIND_8})
+  add_definitions("-DPCURL_SBOX_UNWIND_8")
```
oopsmonk: `add_definitions` is superseded in a newer version and these flags are private flags, in this case should use `target_compile_definitions`.
```suggestion
  target_compile_definitions(common PRIVATE "PCURL_SBOX_UNWIND_8")
```

CMakeLists.txt
```diff
@@ -160,6 +164,23 @@ set(UTILS_SRC
...
+elseif(${PCURL_SBOX_UNWIND_4})
+  add_definitions("-DPCURL_SBOX_UNWIND_4")
+elseif(${PCURL_SBOX_UNWIND_2})
+  add_definitions("-DPCURL_SBOX_UNWIND_2")
+else(${PCURL_SBOX_UNWIND_8})
```
oopsmonk:
```suggestion
else()
```

CMakeLists.txt
```diff
@@ -160,6 +164,23 @@ set(UTILS_SRC
...
+  # Unwind 2 loops by default
+  add_definitions("-DPCURL_SBOX_UNWIND_2")
+endif(${PCURL_SBOX_UNWIND_8})
```
oopsmonk:
```suggestion
endif()
```

CMakeLists.txt
```diff
@@ -160,6 +164,23 @@ set(UTILS_SRC
...
+
+if(${PCURL_STATE_DOUBLE})
+  add_definitions("-DPCURL_STATE_DOUBLE")
+else(${PCURL_STATE_DOUBLE})
```
oopsmonk:
```suggestion
else()
```

CMakeLists.txt
```diff
@@ -160,6 +164,23 @@ set(UTILS_SRC
...
+  add_definitions("-DPCURL_STATE_SHORT")
+endif(${PCURL_STATE_DOUBLE})
```
oopsmonk:
```suggestion
endif()
```

CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ find_package(Threads REQUIRED)
 # default CLI options
 option(CCLIENT "Enable CClient library" ON)
 option(CCLIENT_TEST "Enable CClient library test cases" ON)
+option(PCURL_SBOX_UNWIND_2 "Unwind 2 rounds of PCurl sbox" ON)
```
oopsmonk, line 30: maybe add a comment like:
```suggestion

# PCurl compilation flags
option(PCURL_SBOX_UNWIND_2 "Unwind 2 rounds of PCurl sbox" ON)
```

oopsmonk (changes requested):

CMakeLists.txt
oopsmonk: btw, you have set the default to unwind 2 loops at line 28.
the else case will never be triggered.

CMakeLists.txt
semenov-vladyslav: Done

oopsmonk (dismissed):

thibault-martinez (dismissed):

thibault-martinez (approved):

PR #1421, opened 2019-09-01
ciri: http_destroy - double free

### LINKED ISSUE #1420: CIRI - segmentation fault on iota_api_http_destroy shut down by CTRL-C

tsvisabo:
Fixes #1420


thibault-martinez (changes requested):
How did you come to the conclusion the segmentation fault was due to this ? I don't see a double free here, where would be the other free ? Removing this would cause a leak ?

ciri/api/http/http.c
```diff
@@ -710,7 +710,6 @@ retcode_t iota_api_http_destroy(iota_api_http_t *const api) {
   for (tangle_iter = (tangle_t *)utarray_back(api->tangle_db_connections); tangle_iter != NULL;
        tangle_iter = (tangle_t *)utarray_back(api->tangle_db_connections)) {
     iota_tangle_destroy(tangle_iter);
-    free(tangle_iter);
```
tsvisabo, line 713: @thibault-martinez 
let's avoid commenting on PRs because github won't allow to comment back,
so, the way `utarray` is implemented, it stores all its element inside one memory block
namely, the field `d`, when we call `utarray_free` regardless of whether we defined a DTOR, this `d` field is freed,
also, when you call `utarray_back`, an offset of the `d` block is being returned, and you free that one two, so this is how we got a double free,

hope this makes it clear

ciri/api/http/http.c
```diff
@@ -28,6 +28,10 @@ typedef struct iota_api_http_session_s {
   char_buffer_t *request;
 } iota_api_http_session_t;
 
+
+UT_icd ut_tangle_icd = {sizeof(tangle_t), NULL, NULL, NULL};
+UT_icd ut_spent_addresses_provider_icd = {sizeof(spent_addresses_provider_t), NULL, NULL, NULL};
```
thibault-martinez: Can you make them static ?

thibault-martinez (changes requested):

ciri/api/http/http.c
thibault-martinez: Thanks!

thibault-martinez (approved):

PR #1423, opened 2019-09-02
Milestone tracker refac

tsvisabo:
- Request for branch and trunk of a transaction when it was recently requested
- Clear `requested` set when node is solid
- increase scanning interval

Based on:
https://github.com/iotaledger/iri/pull/1545/files


thibault-martinez:
For the record, I'd like to properly test this before merging it. I'm currently writing a simple tangle cache to check if storage DB I/O was the bottleneck. So I'd like to test this PR + cache as a whole.

thibault-martinez:
@tsvisabo I believe we can close this ?

PR #1445, opened 2019-10-10
utils: bundle miner (WIP)

thibault-martinez:
The bundle miner mines a normalized bundle hash that minimizes the risks of a brute force signature forging attack. It does that by generating a huge amount of bundle hashes with a slight modification in the bundle essence and keeping the one that provide the lowest risk of an attack.


utils/bundle_miner.c
```diff
@@ -0,0 +1,104 @@
...
+#include "utils/bundle_miner.h"
+
+#define OBSOLETE_TAG_OFFSET (243 + 81)
+#define OBSOLETE_TAG_LENGTH 81
+
+/**
+ * Public functions
+ */
+
+double bundle_miner_probability_of_losing(byte_t const *const nb, uint64_t const n) {
```
tsvisabo: This is not the probability of losing because lose is not defined,
anyways, this is the probability of getting out of random selection 81 trytes that each of them has a value smaller than the one currently in _nb_, this probability is of course a metric that correlates with vulnerability but it is not a good metric and it means very little because this event it measures we don't care about because it is super unlikely,
which means that it has the risk of an underflow (_p[1]_*..._p[81]_)
instead, you could count _nb[i]_ and avoid this entire calculation

utils/bundle_miner.c
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return probability;
+}
+
+double bundle_miner_security_level(double const probability, double const radix) {
+  return log(1.0 / probability) / log(radix);
+}
+
+void bundle_miner_min_normalized_bundle(byte_t const *const a, byte_t const *const b, byte_t *const out,
```
tsvisabo: Another option would be to calculate a score for each bundle which is the sum of the values of all trytes and the one with maximal sum get to be chosen, calling this function does not guarantee that the new bundle won't contain a tryte which its value is not related to an already revealed key fragment if that was the original intention,
assume a bundle with only two trytes _min =_ `BA` and a new mined bundle _nb =_ `AZ`, so `out` will be `BZ` which will get a very good score from `bundle_miner_probability_of_losing` so this means _nb_ will be chosen which has `A` with a lower tryte value, meaning an exposed key fragment,
this is ok, you can't expect to mine a bundle without any exposed fragments, but this method is not doing what i suspect it intended to do, plus, what does _out_ is not a real bundle so we should not calculate a score based on it,
we should instead calculate a score for each bundle we mine and the reference one and choose the best

tsvisabo (changes requested):

utils/bundle_miner.h
```diff
@@ -0,0 +1,29 @@
...
+#define __UTILS_BUNDLE_MINER_H__
+
+#include "common/trinary/bytes.h"
+#include "common/trinary/trits.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+double bundle_miner_probability_of_losing(byte_t const *const nb, uint64_t const n);
```
oopsmonk: Documentations

oopsmonk (changes requested):

utils/bundle_miner.h
thibault-martinez: Yeah this is still a very rough draft that I translated from another language. It'll heavily change so I'm not writing documentations yet. But I will! Thanks!

utils/tests/test_bundle_miner.c
```diff
@@ -0,0 +1,121 @@
...
+  tryte_t const* const v1_trytes =
+      (tryte_t const* const) "QVXRKNRXFZIPFPREXRAPNHNSRFFQOWBGCAFZEGFCKDPDXRNVZQ9VJPQPPTFXKPVZVAIENQLETXRVSFKFO";
+  trit_t v1_trits[HASH_LENGTH_TRIT];
+  byte_t nb1[NORMALIZED_BUNDLE_LENGTH];
+  tryte_t const* const v2_trytes =
+      (tryte_t const* const) "JKHLAKTRTDIKMTERIRYEWI9PPOJAKHZEMNCXFB9GTRZRWKSFVAZANHSPABGGQIJAVULKMPPAL9VBSRB9E";
+  trit_t v2_trits[HASH_LENGTH_TRIT];
+  byte_t nb2[NORMALIZED_BUNDLE_LENGTH];
+  byte_t min[NORMALIZED_BUNDLE_LENGTH];
+  size_t const E = 486 * 4;
```
tsvisabo: different name and small letter?

utils/bundle_miner.c
```diff
@@ -0,0 +1,105 @@
...
+    if (TRYTE_VALUE_MAX - a[i] < TRYTE_VALUE_MAX - b[i]) {
+      out[i] = a[i];
+    } else {
+      out[i] = b[i];
+    }
+  }
+}
+
+uint32_t bundle_miner_mine(byte_t const *const min, size_t const number_of_fragments, trit_t *const essence,
+                           size_t const essence_length, uint32_t const offset, uint32_t const count) {
```
tsvisabo: do we need _offset_ param?

tsvisabo (changes requested):

utils/bundle_miner.c
tsvisabo: correction:
this is a good metric but the name is not, this is the probability of getting a random bundle hash which the intermediate hashes needed to sign on are known, having this probability, we can assess what's the expected number of trials an attacker need in order to forge a bundle which he/she knows how to sign on

mobile/android/java/Interface.java
```diff
@@ -8,4 +8,5 @@
     public static native String iota_sign_signature_gen_trytes(String seed, int index, int security, String bundleHash);
     public static native byte[] iota_sign_signature_gen_trits(byte[] seed, int index, int security, byte[] bundleHash);
     public static native String iota_digest(String trytes);
+    public static native String bundle_miner_mine(byte[] bundle_normalized_max, int security, byte[] essence, int essence_length, int count, int nprocs);
```
rajivshah3:
```suggestion
    public static native long bundle_miner_mine(byte[] bundle_normalized_max, int security, byte[] essence, int essence_length, int count, int nprocs);
```
Should this be `long` since the return type in [Interface.h](https://github.com/iotaledger/entangled/pull/1445/files#diff-d3b7c9101f69c7d0205e2b1df1bcda72R71) is `jlong`?

thibault-martinez: Yes, definitely a failed copy/paste, thanks!

utils/bundle_miner.c
thibault-martinez: Removed

utils/bundle_miner.c
```diff
@@ -0,0 +1,157 @@
...
+#include "common/crypto/iss/normalize.h"
+#include "common/crypto/kerl/kerl.h"
+#include "common/defs.h"
+#include "common/trinary/trit_long.h"
+#include "utils/bundle_miner.h"
+#include "utils/handles/thread.h"
+#include "utils/macros.h"
+#include "utils/system.h"
+
+#define OBSOLETE_TAG_OFFSET (243 + 81)
```
oopsmonk:
```suggestion
#define OBSOLETE_TAG_OFFSET (NUM_TRITS_ADDRESS + NUM_TRITS_VALUE)
```

utils/bundle_miner.c
```diff
@@ -0,0 +1,157 @@
...
+#define OBSOLETE_TAG_LENGTH 81
```
oopsmonk: we have `NUM_TRITS_OBSOLETE_TAG`

utils/tests/test_bundle_miner.c
```diff
@@ -0,0 +1,106 @@
...
+  TEST_ASSERT_EQUAL_FLOAT(min_probability, 0.0);
+  TEST_ASSERT_EQUAL_FLOAT(max_probability, 1.0);
+}
+
+static void test_bundle_miner_normalize_hash(void) {
+  tryte_t const* const bundle_trytes =
+      (tryte_t const* const) "QVXRKNRXFZIPFPREXRAPNHNSRFFQOWBGCAFZEGFCKDPDXRNVZQ9VJPQPPTFXKPVZVAIENQLETXRVSFKFO";
+  trit_t bundle_trits[HASH_LENGTH_TRIT];
+  byte_t bundle_expected[NORMALIZED_BUNDLE_LENGTH] = {
+      13, 13, 13, 13, 13, -2, -9, -3, 6,  -1, 9, -11, 6, -11, -9, 5,   -3, -9, 1,  -11, -13, 8,  -13, -8, -9, 6,  6,
```
oopsmonk, line 36: Can you keep the same space number between elements?

oopsmonk (changes requested):
Since it's a new feature/library, Can you descript the problem trying to solve and how do this implementation can fix it?

utils/tests/test_bundle_miner.c
thibault-martinez: No I can't, this is clang-format behavior.

utils/bundle_miner.c
thibault-martinez: Thanks

utils/bundle_miner.c
thibault-martinez: Thanks

thibault-martinez:
@oopsmonk Isn't the documentation enough ?

utils/tests/test_bundle_miner.c
oopsmonk: ah, I didn't notice that sorry.

oopsmonk:
> @oopsmonk Isn't the documentation enough ?

The documentation in headers looks good for me.
I meant having descriptions instead of `No description provided.` in the PR that can help people understand what was this library/feature for. People don't need to dig into the source.

thibault-martinez:
@oopsmonk Yes sure, done. Thanks :)

oopsmonk (approved):

utils/bundle_miner.c
```diff
@@ -0,0 +1,157 @@
...
+  bundle_miner_ctx_t *ctxs = (bundle_miner_ctx_t *)malloc(sizeof(bundle_miner_ctx_t) * procs);
+  thread_handle_t *threads = (thread_handle_t *)malloc(sizeof(thread_handle_t) * procs);
+  uint64_t start_index = 0;
+  double probability = 1.0;
+
+  if (bundle_normalized_max == NULL || essence == NULL || index == NULL) {
+    return RC_NULL_PARAM;
+  }
+
+  if (security > 3 || essence_length % HASH_LENGTH_TRIT != 0) {
```
tsvisabo, line 115: doesn't essence always have the same size?

utils/bundle_miner.c
```diff
@@ -0,0 +1,157 @@
...
+    if (ctxs[i].probability < probability) {
+      probability = ctxs[i].probability;
+      *index = ctxs[i].optimal_index;
+    }
+    free(ctxs[i].essence);
+  }
+
+done:
+
+  free(ctxs);
```
tsvisabo, line 153: what if some of the ctxs have `essence` allocated?

tsvisabo (changes requested):

utils/bundle_miner.c
thibault-martinez: It depends on the bundle size, but this check is only a safety measure to avoid kerl to crash.

tsvisabo (approved):

utils/bundle_miner.c
thibault-martinez: Right

PR #1458, opened 2019-12-11
CClient: update example_node_info.

oopsmonk:
Updated the example of `iota_client_get_node_info`, prints out features and the coordinator address on the console.


JakeSCahill:
When I tested this, I had to replace `get_node_info_req_features_num` with `get_node_info_req_features_len`

oopsmonk:
> When I tested this, I had to replace `get_node_info_req_features_num` with `get_node_info_req_features_len`

yep, `get_node_info_req_features_len` is renamed to `get_node_info_req_features_num` in this PR, so before this PR gets merged, you should use `get_node_info_req_features_len`  in the current repostory.

tsvisabo (approved):

PR #1459, opened 2019-12-16
Update gen_hash_container.sh

MichiKuhl:
The path of the hash_template_dir had a typo.

<<Describe the changes>>
A point was missing in the path. I added it so the file can be executed properly.


oopsmonk:
the gen_hash_container.sh is part of cmake build script, should not be run directly.

PR #1464, opened 2019-12-30
utils/bundle_miner: use mining threshold as a power of 3

tsvisabo:
- Use mining_threshold as a power of 3
- proivding mininig_threshold = 0 will lead to exhaustive mining

### DIFF
--- a/utils/bundle_miner.c
+++ b/utils/bundle_miner.c
@@ -42,6 +42,7 @@ static void *bundle_miner_mine_routine(void *const param) {
   byte_t candidate_normalized[NORMALIZED_BUNDLE_LENGTH];
   byte_t candidate_normalized_max[NORMALIZED_BUNDLE_LENGTH];
   bundle_miner_ctx_t *ctx = (bundle_miner_ctx_t *)param;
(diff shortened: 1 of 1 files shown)


utils/bundle_miner.c
```diff
@@ -61,7 +62,7 @@ static void *bundle_miner_mine_routine(void *const param) {
       if (probability < ctx->probability) {
         ctx->probability = probability;
         ctx->optimal_index = ctx->index;
-        if ((uint32_t)(1 / probability) >= ctx->mining_threshold) {
+        if (num_trials_mining_threshold > 1 && (uint64_t)(1 / probability) >= num_trials_mining_threshold) {
```
thibault-martinez: I think `(1 / probability)` would first be implicitly promoted to something smaller than `uint64_t` then explicitly casted to `uint64_t `; hence, losing some of its value. Just to be sure I think we should suffix `1` with `ULL`.

tsvisabo: great

tsvisabo: Actually, i think it should be 1.0

thibault-martinez (approved):

